PR #15, opened 2019-07-23
Integration of AddressComplete into BaseAddress (#737)

WalterMoar:
Added the AddressComplete code into the BaseComponent. Has some problems that will be fixed as bugs. It will not be used by coops until the bugs are tolerable and some setup is done in the coops app.


vue/sbc-common-components/src/vuelidate.d.ts
```diff
@@ -1,35 +0,0 @@
-//
```
severinbeauvais: Why not just delete this file?

vue/sbc-common-components/docs/BaseAddress/README.md
```diff
@@ -44,14 +44,33 @@ export default class ParentClass extends Vue {
...
+```html
+  <html>
+    <head>
+      <!-- Include the CSS and JS from Canada Post -->
+      <link rel="stylesheet" type="text/css" href="http://ws1.postescanada-canadapost.ca/css/addresscomplete-2.00.min.css"/>
+      <script type="text/javascript" src="http://ws1.postescanada-canadapost.ca/js/addresscomplete-2.00.min.js"></script>
+
+      <script type="text/javascript">
+        // The key for your account.
+        window.addressCompleteKey = 'AB12-CD34-DE56-FG78'
```
severinbeauvais: 'G' is not a valid hex digit :]

vue/sbc-common-components/src/vuelidate.d.ts
WalterMoar: Weird, I did!

vue/sbc-common-components/docs/BaseAddress/README.md
WalterMoar: They're not hex digits, the keyspace would be 6^8*10^8 rather than 26^8*10^8. But I did mess up the sequence!

WalterMoar: Fixed :)

vue/sbc-common-components/src/vuelidate.d.ts
WalterMoar: Oh, the diff page says it was deleted. I think the diff of the lines all gone is just GutHub's way of showing you the contents of the deleted file.

WalterMoar:
> Looks OK.
> Does this component run in the context of another component/app (with the necessary imports)?
> Can this component have unit tests?
> How can I tell that this is working correctly?
> 
> FTR, I'm not necessarily requesting change, just some technical explanations. Or can you think of another place to record that?

This common component is used by both the Directors list and the RegisteredOfficeAddress in the coops-ui. There are tests for it, it wads 100% coverage before the addition of the AddressComplete stuff, but now it's down to 60%. Testing AddressComplete is difficult in that you're integration testing a third party library - I think that would be better covered under an e2e test maybe?

severinbeauvais:
As discussed, this component was tested locally before being submitted into the common components.

Also, I agree that an e2e test would be ideal... Please remember to inform the test guys (or us) to look into this when we get there.

Thanks for everything -- looks good, merging...

PR #34, opened 2019-08-22
[entity1076] Implement Responsive Header

jeznorth:
Implemented responsive header.

- Added vertical logo
- Renamed / standardized naming convention for image assets (gov_bc_logo_vert & gov_bc_logo_horiz)


vue/sbc-common-components/src/components/SbcHeader.vue
```diff
@@ -36,8 +47,7 @@ export default Vue.extend({
           sessionStorage.removeItem('KEYCLOAK_REFRESH_TOKEN')
           sessionStorage.removeItem('KEYCLOAK_TOKEN')
           sessionStorage.removeItem('REGISTRIES_TRACE_ID')
-          sessionStorage.removeItem('AUTH_WEB')
-          window.location.assign('/auth')
+          window.location.assign('/')
```
pksumesh: What is this change for? I guess that /auth was added to take the user to the login page, if it's '/' that may have additional redirection.

jeznorth: Whoa. didn't catch that. Fixing here now.

PR #57, opened 2019-10-02
implemented country drop-down and related UX/UI changes

severinbeauvais:
*Issue #:* /bcgov/entity#1192

*Description of changes:*
- updated my version to 1.0.14
- added packages to get countries and provinces
- changed some v-show to v-if to reduce data loaded multiple times
- now manage addressCountry as 2-char code (in line with schema)
- updated streetAddressAdditional to support more than 1 line
- added missing rules!
- implemented addressRegion as v-select or v-text-field as applicable
- changed country to v-select
- implemented dynamic labels per UX/UI wireframe
- removed confusing/nonfunctional PCA country selector
- changed streetAddressAdditional to include address lines 2, 3, 4 and 5
- allowed null region
- added Province Name to streetAddressAdditional as applicable (for future PCA int'l address lookups)
- implemented mixin to get countries and provinces
- updated and fixed unit tests as applicable

### DIFF
--- a/vue/sbc-common-components/src/components/BaseAddress.vue
+++ b/vue/sbc-common-components/src/components/BaseAddress.vue
@@ -16,14 +16,14 @@
     <!-- Display fields -->
     <v-expand-transition>
       <div class="address-block"
-          v-show="!editing"
+          v-if="!editing"
       >
         <div class="address-block__info">
           <div class="address-block__info-row">
             {{ addressLocal.streetAddress }}
           </div>
-          <div class="address-block__info-row">
-            {{ addressLocal.streetAddressAdditional }}
+          <div class="address-block__info-row pre-wrap"
+            v-html="addressLocal.streetAddressAdditional">
           </div>
           <div class="address-block__info-row">
             <span>{{ addressLocal.addressCity }}</span>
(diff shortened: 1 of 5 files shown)


vue/sbc-common-components/src/components/BaseAddress.vue
```diff
@@ -376,23 +415,42 @@ export default class BaseAddress extends Mixins(ValidationMixin) {
   }
 
   /**
-   * Updates the address data after the user chooses a suggested address.
+   * Callback to update the address data after the user chooses a suggested address.
    *
    * @param address the data object returned by the AddressComplete Retrieve API.
    */
   private addressCompletePopulate (address: object): void {
```
sutherlanda: This is beyond the scope of this PR, but just noticed we are receiving a plain object back from this call (presumably from Canada Post API), and then indexing string properties without any schema check.  I'm wondering if we should create another ticket to make this a bit more robust by validating the returned object first, and if validation fails (because of a schema change), we can handle it a bit more gracefully.  Just a thought, and by no means should that be included here :)

severinbeauvais: That's not a bad idea. The Canada Post "integration" (and I use that word loosely) does not appear to work as documented, so we could likely improve how it functions among other things. I agree this is the subject of another ticket. This ticket was to ensure country was valid (as picked from a list) and I included a number of UI features per the UxPin.

PR #63, opened 2019-10-09
ported all address validation changes from V1

severinbeauvais:
/bcgov/entity#1483

**NOTE: This commit contains changes that coops-ui is not yet ready for -- do not refer to this package version until the parent components that support the new address validation are updated.**

### Description:
- ported all address validation changes from V1 branch
- updated BaseAddress to Vuetify 2
- updated BaseAddress unit tests to Vuetify 2
- updated lint/Babel/Jest configs
- updated package version
- added and updated some packages
- cleaned up prod vs dev dependencies
- downgraded vue test utils due to bug
- misc cleanup


vue/sbc-common-components/src/components/SbcSystemErrorModal.vue
```diff
@@ -6,7 +6,7 @@
 
 <script lang="ts">
 import { Vue, Component, Prop, Emit } from 'vue-property-decorator'
-import SbcSystemError from '@/components/SbcSystemError.vue'
+import SbcSystemError from '../components/SbcSystemError.vue'
```
sutherlanda: Any particular reason we are setting all these to relative paths?  There is a `tsconfig` entry to allow us to use the `@` path...

vue/sbc-common-components/src/views/Home.vue
```diff
@@ -6,7 +6,7 @@
 
 <script lang="ts">
 import { Component, Vue } from 'vue-property-decorator'
-import HelloWorld from '@/components/HelloWorld.vue' // @ is an alias to /src
+import HelloWorld from '../components/HelloWorld.vue'
```
katiemcgoff: Same question as Andrew had above re. relative paths vs. @

vue/sbc-common-components/src/components/SbcSystemErrorModal.vue
severinbeauvais: `@` works fine when the project is run by itself, but when it's a package and its files are imported into a parent project, `@` points to the wrong place. This works around that.

vue/sbc-common-components/src/views/Home.vue
severinbeauvais: Same answer :)

PR #78, opened 2019-11-04
zero fee logic and updated the component to be in ts

saravankumarpa:
Zero fee logic included  for https://github.com/bcgov/entity/issues/1616
Made components to be typescript

### DIFF
--- a/vue/sbc-common-components/src/components/SbcFeeSummary.vue
+++ b/vue/sbc-common-components/src/components/SbcFeeSummary.vue
@@ -5,21 +5,20 @@
     </header>
 
     <div v-show="fetchError">
-      <v-alert
-              :value="true"
-              color="error"
-              icon="warning"
-              outlined
-      >{{fetchError}}
-      </v-alert>
-
+      <v-alert color="error" icon="warning" outlined>{{fetchError}}</v-alert>
     </div>
+
     <v-slide-y-transition group tag="ul" class="fee-list" v-show="!fetchError">
-      <li class="container fee-list__item" v-show="(totalFilingFees > 0 && lineItem.fee) || (totalFilingFees == 0)" v-for="lineItem in fees" :key="lineItem.filingType">
+      <li class="container fee-list__item"
+        v-show="(totalFilingFees > 0 && lineItem.fee) || (totalFilingFees == 0)"
+        v-for="lineItem in fees"
+        :key="lineItem.filingType"
+      >
         <div class="fee-list__item-name">{{lineItem.filingType}}</div>
-        <div class="fee-list__item-value">{{lineItem.fee > 0? (lineItem.fee|currency) : 'No Fee'  }}</div>
+        <div class="fee-list__item-value">{{lineItem.fee > 0 ? (lineItem.fee | currency) : 'No Fee'}}</div>
       </li>
     </v-slide-y-transition>
+
     <div class="container fee-total" v-show="!fetchError">
       <div class="fee-total__name">Total Fees</div>
       <div class="fee-total__currency">CAD</div>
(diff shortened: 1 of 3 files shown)


vue/sbc-common-components/src/components/SbcFeeSummary.vue
```diff
@@ -1,163 +1,152 @@
...
-      }
+             the props accept an array of filing data item's
+                             the fee should have below items
+                             filingDescription : not mandatory.If exists , it will be displayed. Or else what ever from service call will be shown
+                             filingTypeCode : mandatory.Feecodes like OTADD ,OTANN
+                             entityType : for example CP
+
+ */
+    @Prop()
+    filingData: { filingDescription: string, filingTypeCode: string, entityType: string }[]
```
severinbeauvais: Should you assign defaults, as in the old code?

vue/sbc-common-components/src/components/SbcFeeSummary.vue
```diff
@@ -1,163 +1,152 @@
...
-    totalFilingFees () :number {
-      let totalFee = 0
-      if (this.fees) {
-        // console.log('%c FeeMdoule-Watch ZERO FEE %s', 'color: blue ;font-size : 12px', JSON.stringify(this.filingData))
-        totalFee = this.fees.reduce((acc: number, item: { fee: number; }) => acc + item.fee, 0)
-      }
-      this.$emit('total-fee', totalFee)
-      return totalFee
+
+    get totalFilingFees () {
```
severinbeauvais: What type does this getter return?

vue/sbc-common-components/src/components/SbcFeeSummary.vue
```diff
@@ -1,163 +1,152 @@
...
       })
     }
-  }
 }
+
 </script>
 
 <style lang="scss" scoped>
-    @import "../assets/scss/theme.scss";
+  @import "../assets/scss/theme.scss";
```
severinbeauvais: It wouldn't hurt to remove the first level indent for all styles.

vue/sbc-common-components/src/services/fee.services.ts
```diff
@@ -22,6 +23,7 @@ export default {
       promises.push(Axios.get(url, { headers: { Authorization: `Bearer ${token}` } }))
     }
 
+    // @ts-ignore
```
severinbeauvais: Can you look into fixing the statement instead of ignoring the warnings?

vue/sbc-common-components/src/components/SbcFeeSummary.vue
```diff
@@ -1,163 +1,152 @@
...
+      <v-alert
+              :value="true"
+              color="error"
+              icon="warning"
+              outlined
+      >{{fetchError}}
+      </v-alert>
+
+    </div>
+    <v-slide-y-transition group tag="ul" class="fee-list" v-show="!fetchError">
```
severinbeauvais: Although this was existing code, I recommend removing the v-slide-y-transition as this animation looks really weird when you Cease and then Undo a director.

vue/sbc-common-components/src/components/SbcFeeSummary.vue
```diff
@@ -1,163 +1,152 @@
...
-    filingData: function (newVal) {
-      // console.log('%c FeeMdoule-Watch Activated as %s', 'color: blue ;font-size : 12px', JSON.stringify(this.filingData))
+
+    @Watch('filingData')
+    onFilingDataChanged (val: string, oldVal: string) {
+      console.log('%c FeeMdoule-Watch Activated as %s', 'color: blue ;font-size : 12px', JSON.stringify(this.filingData))
       FeeServices.getFee(this.filingData, this.payURL).then((data: any) => {
         this.fetchError = ''
         this.fees = data
+        this.$emit('total-fee', this.totalFilingFees)
```
severinbeauvais: There is a more standard TS way to emit an event. Have a look at BaseAddress.vue.

vue/sbc-common-components/src/components/SbcFeeSummary.vue
```diff
@@ -1,163 +1,152 @@
...
-            <div class="fee-total__name">Total Fees</div>
-            <div class="fee-total__currency">CAD</div>
-            <div class="fee-total__value">
-                <v-slide-y-reverse-transition name="slide" mode="out-in">
-                    <div>{{totalFilingFees | currency}}</div>
-                </v-slide-y-reverse-transition>
-            </div>
-        </div>
-    </v-card>
+  <v-card>
```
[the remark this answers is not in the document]
severinbeauvais: This is warning, not an error, but it should be fixed (in a different ticket). It happens because, in the template v-for loop, you use `:key="lineItem.filingType`, but it's possible for 2 items to have the same filing type (eg, a free dir change and a paid dir change).

vue/sbc-common-components/src/components/SbcFeeSummary.vue
severinbeauvais: Does not reproduce for me atm.

vue/sbc-common-components/src/components/SbcFeeSummary.vue
severinbeauvais: Fixed in next commit.

vue/sbc-common-components/src/components/SbcFeeSummary.vue
severinbeauvais: Fixed in next commit.

vue/sbc-common-components/src/components/SbcFeeSummary.vue
severinbeauvais: Fixed in next commit.

vue/sbc-common-components/src/components/SbcFeeSummary.vue
severinbeauvais: Fixed in next commit.

vue/sbc-common-components/src/services/fee.services.ts
severinbeauvais: Fixed in next commit.
The cause was that, in case of a catch, "void" was being returned. I solved this (not by updating the return type to `Promise<void | Fee[]>` but instead) by returning `[]` in case of an error. This satisfies the typing.

vue/sbc-common-components/src/components/SbcFeeSummary.vue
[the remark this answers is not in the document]
saravankumarpa: https://github.com/bcgov/entity/issues/1657

PR #79, opened 2019-11-07
make Province and Country appear disabled

severinbeauvais:
*Issue #:* /bcgov/entity/1673

*Description of changes:*
- made readonly inputs look disabled (while still displaying normal error styling)
- updated prod vs dev dependencies

**Note that this is not a high-priority fix so I did not update the package version number. I am aware that other things needs to be fixed (ie, icons) in this branch, so when those are fixed then we can update the version.**

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the lear license (Apache 2.0).

### DIFF
--- a/vue/sbc-common-components/package.json
+++ b/vue/sbc-common-components/package.json
@@ -19,7 +19,6 @@
     "docs:build": "vuepress build docs"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


severinbeauvais:
Approved by Andrew so I'm merging this before the changes get lost.

PR #88, opened 2019-12-10
Keycloak service to support refreshing of the tokens

saravankumarpa:


severinbeauvais:
@saravankumarpa Can you check in a test/sample component that shows how to use this? Does it need to be registered? Installed? (Mixin? Vue.use()?) Run periodically?

saravankumarpa:
> @saravankumarpa Can you check in a test/sample component that shows how to use this? Does it need to be registered? Installed? (Mixin? Vue.use()?) Run periodically?

gonna add vuepress docs soon..

vue/sbc-common-components/src/services/token.services.ts
```diff
@@ -0,0 +1,119 @@
...
+  private syncSessionStorage () {
+    sessionStorage.setItem('KEYCLOAK_TOKEN', this.kc.token)
+    sessionStorage.setItem('KEYCLOAK_REFRESH_TOKEN', this.kc.refreshToken)
+    sessionStorage.setItem('ID_TOKEN', this.kc.idToken)
+  }
+
+  decodeToken () {
+    try {
+      let token = sessionStorage.getItem('KEYCLOAK_TOKEN')
+      const base64Url = token.split('.')[1]
```
severinbeauvais: Should check that token != null before splitting it or this will crash. Are you guaranteed that session storage has the token?

vue/sbc-common-components/docs/TokenService/README.md
```diff
@@ -0,0 +1,55 @@
...
+        `Vue.prototype.$tokenService = new TokenService()`
+   
+   2. Init the tokenservice so that keycloak service is being set up
+   3. start the timer
+    
+        ```if (sessionStorage.getItem('KEYCLOAK_TOKEN')) {
+            // eslint-disable-next-line no-console
+            console.info('[APP.vue] Token exists.So start the refreshtimer')
+            var self = this
+            this.$tokenService.initUsingUrl(`/${process.env.VUE_APP_PATH}/config/kc/keycloak.json`).then(function (success) {
```
severinbeauvais: Can you please attach this JSON file to this PR?

saravankumarpa: is it for local testing u mean?

vue/sbc-common-components/src/services/token.services.ts
saravankumarpa: Thanks.Will add it

vue/sbc-common-components/src/services/token.services.ts
```diff
@@ -0,0 +1,119 @@
...
+    let self = this
+
+    // check if refresh token is still valid . Or else clear all timers and throw errors
+    let refreshTokenExpiresIn = self.kc.refreshToken['exp'] - Math.ceil(new Date().getTime() / 1000) + self.kc.timeSkew
+    if (refreshTokenExpiresIn < 0) {
+      throw new Error('Refresh Token Expired..No more token refreshes')
+    }
+
+    let expiresIn = self.kc.tokenParsed['exp'] - Math.ceil(new Date().getTime() / 1000) + self.kc.timeSkew
+    let refreshInMilliSeconds = (expiresIn * 1000) - 2000 // refresh before two second of expiry
```
severinbeauvais: 2 seconds seems very short... it doesn't give much time to refresh the token. The network may be congested, and the in-browser timer may slow down for a lot of reasons:
- the CPU is overloaded
- the browser tab is in the background mode
- the laptop is on battery

Is there a recommended value for this? 30 seconds maybe?

Ref: https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout#Reasons_for_delays_longer_than_specified

vue/sbc-common-components/docs/TokenService/README.md
```diff
@@ -0,0 +1,55 @@
...
+The abstraction is provided so that COOPS-UI project or any other consumer of auth-web can stay independent of keycloak logic.
+
+
+It supports two modes for **refreshing the token**
+
+
+
+### Timer Mode
+   
+   In this case , it runs a timer [a plain old javascript settimeout] . It calculates the expiry of current keycloak token and schedules a timeout before 2 seconds of expiry
```
severinbeauvais: NB: setTimeout() isn't a JavaScript function. It's a browser function :]

vue/sbc-common-components/docs/TokenService/README.md
```diff
@@ -0,0 +1,55 @@
...
+   Since the timers get destroyed on page refresh , execute the logic in start up or when user signs in
+   
+   a sample usage
+   
+   1. Register the compoent in vue
+        
+        `Vue.prototype.$tokenService = new TokenService()`
```
severinbeauvais: Should be `new TokenServices()`. Also below.

vue/sbc-common-components/src/services/token.services.ts
saravankumarpa: Will add it as a config may be... 
30 seconds is too much..  this is the time  before expiry when we are refreshing..so 5 seconds or 10 seconds is good enough

vue/sbc-common-components/src/services/token.services.ts
```diff
@@ -0,0 +1,119 @@
...
+    return new Promise((resolve, reject) => {
+      this.kc = Keycloak(keyCloakConfigurl)
+      this.kc.init(this.kcOptions
+      ).success(function (authenticated) {
+        console.info('[TokenServices] is User Authenticated:Syncing' + authenticated)
+        self.syncSessionStorage()
+        resolve(self.kc.token)
+      }).error(function (err) {
+        console.info('[TokenServices] Fatal Error:Cudnt Initialise KC instance' + err)
+        reject(new Error('Cudnt Initialise KC'))
```
severinbeauvais: Please use real language in the error messages. Also on line 58.

vue/sbc-common-components/docs/TokenService/README.md
severinbeauvais: Yes it was for local testing, but I got a copy separately.

vue/sbc-common-components/src/services/token.services.ts
severinbeauvais: Another reason why 2 seconds is probably too short:

skew

vue/sbc-common-components/src/services/token.services.ts
```diff
@@ -0,0 +1,119 @@
...
+    onLoad: 'login-required',
+    checkLoginIframe: false,
+    timeSkew: 0,
+    token: sessionStorage.getItem('KEYCLOAK_TOKEN'),
+    refreshToken: sessionStorage.getItem('KEYCLOAK_REFRESH_TOKEN'),
+    idToken: sessionStorage.getItem('KEYCLOAK_ID_TOKEN')
+  }
+
+  initUsingKc (kcInstance: KeycloakInstance) {
+    console.info('[TokenServices] KC Instance registered using kc instance')
```
severinbeauvais: How about providing an option to disable all info messages? This module is very noisy :)

vue/sbc-common-components/src/services/token.services.ts
```diff
@@ -0,0 +1,119 @@
...
+  private kcLoaded
+  private counter = 0
+  private timerId
+  private kcOptions: KeycloakInitOptions = {
+    onLoad: 'login-required',
+    checkLoginIframe: false,
+    timeSkew: 0,
+    token: sessionStorage.getItem('KEYCLOAK_TOKEN'),
+    refreshToken: sessionStorage.getItem('KEYCLOAK_REFRESH_TOKEN'),
+    idToken: sessionStorage.getItem('KEYCLOAK_ID_TOKEN')
```
severinbeauvais: Should all 3 of these be in session storage before initializing kc on line 28?

vue/sbc-common-components/src/services/token.services.ts
```diff
@@ -0,0 +1,119 @@
...
+  stopRefreshTimer () {
+    console.info('[TokenServices Stopping the timer] ')
+    clearTimeout(this.timerId)
+  }
+
+  private scheduleRefreshToken () {
+    let self = this
+
+    // check if refresh token is still valid . Or else clear all timers and throw errors
+    let refreshTokenExpiresIn = self.kc.refreshToken['exp'] - Math.ceil(new Date().getTime() / 1000) + self.kc.timeSkew
```
severinbeauvais: `self.kc.refreshToken['exp']` is initially undefined, so `refreshTokenExpiresIn` is NaN.

Lucky for you `NaN < 0` is false so the next block of code is not executed.

But there is a logic problem here that should be fixed.

vue/sbc-common-components/src/services/token.services.ts
saravankumarpa: the whole tokenservice is invalid when the session storage doesnt have token.. so may be I will add a check on init to address that

vue/sbc-common-components/src/services/token.services.ts
saravankumarpa: Yup..5 0r 10 is good..

vue/sbc-common-components/src/services/token.services.ts
saravankumarpa: Sure Thing.I will add it

vue/sbc-common-components/src/services/token.services.ts
saravankumarpa: self.kc.refreshToken['exp'] shouldnt be Nan when scheduleRefreshToken is invoked. the no token check will be in init. The the ['exp'] will never be undefined.

saravankumarpa: self.kc.refreshToken['exp'] shouldnt be Nan when scheduleRefreshToken is invoked. the no token check will be in init. The the ['exp'] will never be undefined
Whats the logic problem?

severinbeauvais: I found that when I schedule a timer refresh (right after first init), kc does not contain property 'exp' (it's undefined). So the math above fails with NaN.

Am I missing an init step? I just copied the code in the readme.

severinbeauvais: I found that when I schedule a timer refresh (right after first init), kc does not contain property 'exp' (it's undefined). So the math above fails with NaN.

Am I missing an init step? I just copied the code in the readme.

saravankumarpa: I fixed the wrong variable ,thanks for noticing it

saravankumarpa:
@severinbeauvais the left over comments are already addressed and will be in a new PR.. merging this now for the sake of deploying into DEV

PR #94, opened 2019-12-20
prevent extra whitespace in addresses

severinbeauvais:
- added whitespace rules to all inputs
- replaced invalid 'name' attributes with classnames
- replaced lodash methods with ES6 ones
- removed original address that didn't work ('modified' event was never used anyway)
- fixed AC update that sometimes didn't validate
- initialized Vuetify rules object to prevent template errors
- general simplification and cleanup
- fixed unit tests
- removed lodash dependency
- updated version number

### DIFF
--- a/vue/sbc-common-components/package.json
+++ b/vue/sbc-common-components/package.json
@@ -1,6 +1,6 @@
 {
   "name": "sbc-common-components",
-  "version": "2.0.29",
+  "version": "2.0.30",
   "private": false,
   "description": "Common Vue Components to be used across SBC projects",
   "license": "Apache-2.0",
@@ -19,12 +19,10 @@
     "docs:build": "vuepress build docs"
   },
   "dependencies": {
+    "@mdi/font": "^4.5.95",
     "axios": "^0.18.0",
     "core-js": "^3.1.4",
     "country-list": "^2.2.0",
-    "lodash.set": "^4.3.2",
-    "lodash.unset": "^4.5.2",
-    "@mdi/font": "^4.5.95",
     "provinces": "^1.11.0",
     "regenerator-runtime": "^0.13.3",
     "vue": "^2.6.10",
--- a/vue/sbc-common-components/src/components/BaseAddress.vue
+++ b/vue/sbc-common-components/src/components/BaseAddress.vue
@@ -12,124 +12,107 @@
 //
 
 <template>
-  <div class="meta-container__inner">
+  <div class="base-address">
     <!-- Display fields -->
     <v-expand-transition>
-      <div class="address-block"
-          v-if="!editing"
-      >
-        <div class="address-block__info">
-          <div class="address-block__info-row">
-            {{ addressLocal.streetAddress }}
-          </div>
-          <div class="address-block__info-row pre-wrap"
-            v-html="addressLocal.streetAddressAdditional">
-          </div>
+      <div v-if="!editing" class="address-block">
+        <div class="address-block__info pre-wrap">
+          <div class="address-block__info-row">{{ addressLocal.streetAddress }}</div>
+          <div class="address-block__info-row">{{ addressLocal.streetAddressAdditional }}</div>
           <div class="address-block__info-row">
             <span>{{ addressLocal.addressCity }}</span>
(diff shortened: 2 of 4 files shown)


vue/sbc-common-components/src/components/BaseAddress.vue
```diff
@@ -12,124 +12,107 @@
...
-            <span v-if="addressLocal.postalCode !== 'N/A'">&nbsp;&nbsp;{{ addressLocal.postalCode }}</span>
-          </div>
-          <div class="address-block__info-row">
-            {{ getCountryName(addressCountry) }}
-          </div>
-          <div class="address-block__info-row"
-               v-if="addressLocal.deliveryInstructions"
-          >
-            {{ addressLocal.deliveryInstructions }}
+            <span v-if="addressLocal.addressRegion">&nbsp;{{ addressLocal.addressRegion }}</span>
```
saravankumarpa: may be just add a comment why nbsp is used?  is it for spacing needs or some other purpose?

vue/sbc-common-components/src/components/BaseAddress.vue
```diff
@@ -291,6 +261,15 @@ export default class BaseAddress extends Mixins(ValidationMixin, CountriesProvin
     return Boolean(this.schemaLocal && this.schemaLocal[prop] && this.schemaLocal[prop].isBC)
   }
 
+  /**
```
saravankumarpa: i have no clue on the purpose..so a shot in dark... https://vuejs.org/v2/guide/forms.html#trim does it help in anyway?

vue/sbc-common-components/package.json
```diff
@@ -19,12 +19,10 @@
     "docs:build": "vuepress build docs"
   },
   "dependencies": {
+    "@mdi/font": "^4.5.95",
     "axios": "^0.18.0",
     "core-js": "^3.1.4",
     "country-list": "^2.2.0",
-    "lodash.set": "^4.3.2",
```
saravankumarpa: thats great

vue/sbc-common-components/src/components/BaseAddress.vue
```diff
@@ -291,6 +261,15 @@ export default class BaseAddress extends Mixins(ValidationMixin, CountriesProvin
...
+   * Array of validation rules used by input elements to prevent extra whitespace.
+   */
+  private readonly spaceRules: Array<Function> = [
+    v => !/^\s/g.test(v) || 'Invalid spaces', // leading spaces
```
saravankumarpa: I will try to handle leading and trailing spaces in one test statement ;just to avoid the two tests

vue/sbc-common-components/src/components/BaseAddress.vue
severinbeauvais: This is a non-breaking space and is the proper and usual way to provide real white-space in text in HTML. Often developers "assume" that there will be a space provided when elements have a space character between them or are on different lines ... this is not always true, depending on the type of element and the white-space CSS property. In this case I was explicit about providing a whitespace character in front of this span element. I don't think it needs a comment: I'm just doing something that all devs should do in this circumstance. Unless you think I should document this for junior HTML devs that haven't seen this before? In that case, let me know and I will add it in a future commit.

vue/sbc-common-components/src/components/BaseAddress.vue
severinbeauvais: Good question. It sometimes works, sometimes for leading spaces and less often for trailing spaces and never for spaces between words. But it doesn's always work, while using a rule and a message always works and additionally shows the user the problem in case they edit an existing address with too much whitespace.

vue/sbc-common-components/src/components/BaseAddress.vue
severinbeauvais: Yes, I could combine the 2 statements, but I think this is much easier to understand (and review) and for that reason I'd like to leave it as-is. I hope you're ok with that :)

severinbeauvais:
I had 2 reviewers and 2 approvals so I went ahead and merged this. Any outstanding code changes can be made in a future commit (they are not functional).

PR #95, opened 2020-01-06
made Tokenservice stricter

saravankumarpa:
Changes for making tokenservice strict

List of changes
- used explicit null checks instead of non-null assertion operator in most of the places.Can use non-null assertion operator ;but sonar will throw errors if used

- removed some console.logs to reduce noise

### DIFF
--- a/vue/sbc-common-components/src/services/token.services.ts
+++ b/vue/sbc-common-components/src/services/token.services.ts
@@ -1,45 +1,46 @@
 import Keycloak, { KeycloakInitOptions, KeycloakInstance } from 'keycloak-js'
 
 class TokenServices {
-  private kc: KeycloakInstance
+  private kc: KeycloakInstance | undefined
   private static instance: TokenServices
-  private kcLoaded
   private counter = 0
   private REFRESH_ATTEMPT_INTERVAL = 10 // in seconds
-  private timerId
-  private kcOptions: KeycloakInitOptions = {
-    onLoad: 'login-required',
-    checkLoginIframe: false,
-    timeSkew: 0,
-    token: sessionStorage.getItem('KEYCLOAK_TOKEN'),
-    refreshToken: sessionStorage.getItem('KEYCLOAK_REFRESH_TOKEN'),
-    idToken: sessionStorage.getItem('KEYCLOAK_ID_TOKEN')
-  }
+  private timerId = 0
 
   initUsingKc (kcInstance: KeycloakInstance) {
-    console.info('[TokenServices] KC Instance registered using kc instance')
     this.kc = kcInstance
   }
 
   initUsingUrl (keyCloakConfigurl: string) {
-    console.info('[TokenServices] KC Instance registered using kc URL')
     var self = this
+
+    const kcOptions: KeycloakInitOptions = {
+      onLoad: 'login-required',
+      checkLoginIframe: false,
+      timeSkew: 0,
+      token: sessionStorage.getItem('KEYCLOAK_TOKEN') || undefined,
+      refreshToken: sessionStorage.getItem('KEYCLOAK_REFRESH_TOKEN') || undefined,
+      idToken: sessionStorage.getItem('KEYCLOAK_ID_TOKEN') || undefined
+    }
+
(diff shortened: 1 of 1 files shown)


vue/sbc-common-components/src/services/token.services.ts
```diff
@@ -69,20 +74,26 @@ class TokenServices {
...
+      refreshTokenExpiresIn = self.kc.refreshTokenParsed['exp']! - Math.ceil(new Date().getTime() / 1000) + self.kc.timeSkew
+    }
     if (refreshTokenExpiresIn < 0) {
       throw new Error('Refresh Token Expired..No more token refreshes')
     }
-
-    let expiresIn = self.kc.tokenParsed['exp'] - Math.ceil(new Date().getTime() / 1000) + self.kc.timeSkew
+    let expiresIn = -1
+    if (self.kc !== undefined && self.kc.tokenParsed !== undefined && self.kc.tokenParsed['exp'] !== undefined) {
+      expiresIn = self.kc.tokenParsed['exp'] - Math.ceil(new Date().getTime() / 1000) + self.kc!.timeSkew!
```
sutherlanda: Do you need `self.kc!` non-null assertion operator here?  Your condition on line 86 should be sufficient.

vue/sbc-common-components/src/services/token.services.ts
```diff
@@ -92,26 +103,36 @@ class TokenServices {
...
     }, refreshInMilliSeconds)
-    console.log('[TokenServices] Timer id:' + this.timerId)
   }
 
   private syncSessionStorage () {
-    sessionStorage.setItem('KEYCLOAK_TOKEN', this.kc.token)
-    sessionStorage.setItem('KEYCLOAK_REFRESH_TOKEN', this.kc.refreshToken)
-    sessionStorage.setItem('KEYCLOAK_ID_TOKEN', this.kc.idToken)
+    if (this.kc) {
+      if (this.kc.token != null) {
```
sutherlanda: Do you need to explicitly check for equality to null? Maybe better just to do `if (this.kc.token) {`

vue/sbc-common-components/src/services/token.services.ts
```diff
@@ -92,26 +103,36 @@ class TokenServices {
...
   decodeToken () {
     try {
       let token = sessionStorage.getItem('KEYCLOAK_TOKEN')
-      const base64Url = token.split('.')[1]
-      const base64 = decodeURIComponent(window.atob(base64Url).split('').map(function (c) {
-        return '%' + ('00' + c.charCodeAt(0).toString(16)).slice(-2)
-      }).join(''))
-      return JSON.parse(base64)
+      if (token != null) {
+        const base64Url = token.split('.')[1]
```
sutherlanda: This will throw a runtime exception if the token is invalid.  Maybe an additional check here.

vue/sbc-common-components/src/services/token.services.ts
saravankumarpa: hm..the kc.token is optional param ..so have to check this explicitly... and since kc is not initialised in constructor , it will warn about object possibly null

vue/sbc-common-components/src/services/token.services.ts
saravankumarpa: done

vue/sbc-common-components/src/services/token.services.ts
saravankumarpa: totally unused function.. I might as well just remove it

vue/sbc-common-components/src/services/token.services.ts
severinbeauvais: Lint will complain about the existing expression because you are using `!=` instead of `!==`.

You may intend the first one and the second one may break your logic. The difference is the first will compare "nullish" (equality) whereas the second will compare type (strict equality).

https://stackoverflow.com/questions/16204840/comparing-to-null-vs-in-javascript
http://adripofjavascript.com/blog/drips/equals-equals-null-in-javascript.html

PR #99, opened 2020-01-23
#2093 - Changes to support account ui dropdown in header

sutherlanda:
Please squash before merging


vue/sbc-common-components/src/components/SbcHeader.vue
```diff
@@ -86,6 +139,14 @@ export default class SbcHeader extends Vue {
   goToUserProfile () {
     window.location.assign('/cooperatives/auth/userprofile')
   }
+
+  goToAccountSettings () {
```
saravankumarpa: may be in another ticket , we have to implement the logic of checking if the user is in auth application , we can emit event or else we can assign the location ; so that the page wont be refreshed if user is in auth itself

PR #100, opened 2020-01-24
#2439 - Configure sbc-common-components LaunchDarkly for feature flags

sutherlanda:
This PR adds capability to load LaunchDarkly flagsets and syncs with session storage.


vue/sbc-common-components/src/components/SbcHeader.vue
```diff
@@ -107,44 +108,90 @@
...
-  get accountType (): string {
-    return sessionStorage.getItem('USER_ACCOUNT_TYPE') || 'BCSC'
+  private get accountType (): string {
+    return ConfigHelper.getFromSession(SessionStorageKeys.UserAccountType) || 'BCSC'
+  }
+
+  private mounted () {
+    // Initialize LaunchDarkly flags and sync to session storage
+    const user = { 'key': 'sbc-common-components' }
+    this.ldClient = initialize('5db9da115f58e008123cd783', user)
```
sutherlanda: Just a note for reviewers.  This client id for the LaunchDarkly project is not secret, and I am hardcoding it until we finish off the endpoint to return header configuration from the app (it will be included in that payload).

vue/sbc-common-components/src/services/token.services.ts
```diff
@@ -77,14 +78,14 @@ class TokenServices {
   private scheduleRefreshToken (refreshEarlyTimeinMilliseconds: number) {
     let refreshTokenExpiresIn = -1
     // check if refresh token is still valid . Or else clear all timers and throw errors
-    if (this.kc && this.kc.timeSkew && this.kc.refreshTokenParsed) {
+    if (this.kc && this.kc.timeSkew !== undefined && this.kc.refreshTokenParsed) {
```
severinbeauvais: I had to think here about why you did this, then realized that a value of 0 is valid.

vue/sbc-common-components/src/services/token.services.ts
```diff
@@ -111,21 +112,21 @@ class TokenServices {
...
+        ConfigHelper.addToSession(SessionStorageKeys.KeyCloakIdToken, this.kc.idToken)
       }
     }
   }
 
   decodeToken () {
     try {
-      let token = sessionStorage.getItem('KEYCLOAK_TOKEN')
-      if (token != null) {
+      let token = ConfigHelper.getFromSession(SessionStorageKeys.KeyCloakToken)
```
severinbeauvais: Use `const` when property isn't reassigned later. This will avoid some code debt for when `strict=true` is set (or required by apps that use this).

vue/sbc-common-components/src/util/config-helper.ts
```diff
@@ -0,0 +1,17 @@
+export default class ConfigHelper {
```
severinbeauvais: If you changed this to a mixin then you wouldn't have to call each method as "ConfigHelper.xxx()" 😁 

But I DON'T know how to use a mixin in a non-Vue class 😒

vue/sbc-common-components/src/util/constants.ts
```diff
@@ -0,0 +1,13 @@
+export enum SessionStorageKeys {
+  KeyCloakToken = 'KEYCLOAK_TOKEN',
```
severinbeauvais: BTW, Keycloak is 1 word (not cap C). I know, I know, it's abbreviated as KC 🤷‍♂

vue/sbc-common-components/src/components/SbcHeader.vue
severinbeauvais: Did you know about package "ld-vue"? It may simplify things for you.

sutherlanda: I did investigate that, but it requires a Vue instance to attach to.

vue/sbc-common-components/src/components/SbcHeader.vue
```diff
@@ -107,44 +108,90 @@
...
+    this.ldClient.on('ready', () => {
+      console.log('It\'s now safe to request feature flags')
+      ConfigHelper.addToSession(SessionStorageKeys.LaunchDarklyFlags, JSON.stringify(this.ldClient.allFlags()))
+    })
   }
 
-  logout () {
+  private logout () {
```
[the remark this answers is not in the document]
sutherlanda: This routine should only be called as the event handler on the Logout button in the markup, hence why it is marked private.  You are calling this `logout` routine from a dialog in Coops?

vue/sbc-common-components/src/services/token.services.ts
sutherlanda: That's correct.

vue/sbc-common-components/src/services/token.services.ts
sutherlanda: Changed.  BTW this has nothing to do with strict mode.

vue/sbc-common-components/src/util/config-helper.ts
sutherlanda: Might be nice, but not doing this for now.

vue/sbc-common-components/src/util/constants.ts
sutherlanda: Thanks.  Leaving as is for now.

vue/sbc-common-components/src/components/SbcHeader.vue
```diff
@@ -107,44 +108,90 @@
...
     window.location.assign('/cooperatives/auth/signout')
   }
 
-  login () {
+  private login () {
     window.location.assign('/cooperatives/auth/signin/bcsc')
   }
 
-  goToUserProfile () {
+  private goToUserProfile () {
```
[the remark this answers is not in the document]
sutherlanda: Not sure what this comment has to do with marking the routine as private.  The intent is to prevent the routines from called from outside the component.

vue/sbc-common-components/src/components/SbcHeader.vue
severinbeauvais: Oh, OK, nvm!

vue/sbc-common-components/src/components/SbcHeader.vue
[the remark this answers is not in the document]
severinbeauvais: Nothing with making this private. Sorry, was just curious.

vue/sbc-common-components/src/util/config-helper.ts
severinbeauvais: np

vue/sbc-common-components/src/components/SbcHeader.vue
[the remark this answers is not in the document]
severinbeauvais: Yes, we're calling logout() using the component ref. I'm open to alternatives.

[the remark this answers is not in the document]
sutherlanda: Ah okay, sorry did not realize.  Ideally we'd have a service call for logging out that you would call, but since we don't have that, I'll unmark this as private.  Thanks for catching that.

PR #104, opened 2020-01-29
reverted TS optional chaining due to builder error

severinbeauvais:
*Issue #:* /bcgov/entity#2357

*Description of changes:*
- reverted TS optional chaining due to builder error
- did not update version number -- this can go in with next fix 

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the lear license (Apache 2.0).

### DIFF
--- a/vue/sbc-common-components/src/services/fee.services.ts
+++ b/vue/sbc-common-components/src/services/fee.services.ts
@@ -31,9 +31,9 @@ export default {
         return args
           .map(response => response.data as PayData)
           .map(data => {
+            const filingDatum = filingData.find(fd => fd.filingTypeCode === data.filingTypeCode)
             // default the title if client hasn't passed this on
-            const filingDescription = filingData.find(fd => fd.filingTypeCode === data.filingTypeCode)?.filingDescription
-            const filingType = filingDescription || data.filingType
+            const filingType = (filingDatum && filingDatum.filingDescription) ? filingDatum.filingDescription : data.filingType
             // total fees is a sum of filingFees,serviceFees,processingFees , gst , pst
             const fee = data.filingFees + data.serviceFees + data.processingFees + data.tax.gst + data.tax.pst
             return { fee, filingType } as Fee


sutherlanda:
@severinbeauvais what was the build error? Was hoping to use optional chaining in the next set of changes, but I won't if it's causing issues.

severinbeauvais:
> @severinbeauvais what was the build error? Was hoping to use optional chaining in the next set of changes, but I won't if it's causing issues.

This is the build error when I used optional chaining. I didn't notice it when doing `npm run build` in THIS project (which uses TS 3.7.3) but I saw it in the bcrs-business-create-ui project (which uses TS 3.5.3). So maybe we just need to update our TS version, but that can wait for another day and THIS solution is adequate for now (imho).

build error

severinbeauvais:
PS Yes, optional chaining is new in TS 3.7.
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html

sutherlanda:
Okay yeah if any clients are using below 3.7 for TypeScript we shouldn't use optional chaining.  I will need to update my PR as well.

Is upgrading a bcrs-business-create-ui to 3.7 a task that could be completed in the short term?

severinbeauvais:
> Okay yeah if any clients are using below 3.7 for TypeScript we shouldn't use optional chaining. I will need to update my PR as well.
> 
> Is upgrading a bcrs-business-create-ui to 3.7 a task that could be completed in the short term?

I'm not sure. I know that in coops-ui we have fixed the Typescript version to "<3.6.0" due to issues with other dependencies, but that may all be resolved now.

The issue with upgrading both projects is that we'll have to do basic regression testing, so I created ticket /bcgov/entity#2517 to track that. Let me know if you think this needs to be made high priority.

sutherlanda:
Sounds good - I don't think it's high priority.  I'll remove the chaining from my PR for now.  Thanks for creating the ticket!

PR #107, opened 2020-02-11
Display Priority, Future Effective and Transaction Fees in Fee Module

nitheesh-aot:
This PR is for display Priority, Future Effective and Transaction Fees in Fee Module.

Issue #: 2604
https://github.com/bcgov/entity/issues/2604

### DIFF
--- a/vue/sbc-common-components/src/components/SbcFeeSummary.vue
+++ b/vue/sbc-common-components/src/components/SbcFeeSummary.vue
@@ -25,8 +25,7 @@
           :key="lineItem.filingType+'-priority'"
           >
           <div class="fee-list__item-name">Priority Fee</div>
-          <!-- <div class="fee-list__item-value">{{lineItem.priorityFees | currency}}</div> -->
-          <div class="fee-list__item-value">{{lineItem.priorityFees}}</div>
+          <div class="fee-list__item-value">{{lineItem.priorityFees | currency}}</div>
         </li>
         <li class="container fee-list__item"
           v-if="lineItem.futureEffectiveFees"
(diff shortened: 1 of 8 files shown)


sutherlanda:
@nitheesh-aot don't forget to increment the version number in the `package.json` or the GH action for publish will fail.

nitheesh-aot:
> @nitheesh-aot don't forget to increment the version number in the `package.json` or the GH action for publish will fail.

Will change the version, Thanks @sutherlanda

vue/sbc-common-components/src/services/fee.services.ts
```diff
@@ -21,7 +21,7 @@ export default {
       if (!filing.filingTypeCode) {
         Promise.resolve()
       }
-      const url = `${payApiUrl}fees/${filing.entityType}/${filing.filingTypeCode}?waiveFees=${!!filing.waiveFees}`
+      const url = `${payApiUrl}fees/${filing.entityType}/${filing.filingTypeCode}?waiveFees=${!!filing.waiveFees}&priority=${!!priority}&futureEffective=${!!futureEffective}`
```
saravankumarpa: may be send it only if its true...

nitheesh-aot: Thanks @saravankumarpa, I have corrected this, please do re-review

vue/sbc-common-components/src/components/SbcFeeSummary.vue
```diff
@@ -47,6 +71,12 @@ export default class SbcFeeSummary extends Vue {
   @Prop({ default: '' })
   private payURL!: string
 
+  @Prop({ default: false })
+  private priority!: boolean
```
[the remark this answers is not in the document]
nitheesh-aot: checked with @pksumesh,
since the query param name in the backend is the same, we can keep it this way.

also its now moved to filingData object rather than a separate prop.

PR #119, opened 2020-02-26
context path from application router and updated doc

nitheesh-aot:
- taking context path from the application router.
- updated doc with the route config examples

### DIFF
--- a/vue/sbc-common-components/src/components/SbcHeader.vue
+++ b/vue/sbc-common-components/src/components/SbcHeader.vue
@@ -231,9 +231,9 @@ export default class SbcHeader extends NavigationMixin {
   logout () {
     if (this.redirectOnLogout) {
       const url = encodeURIComponent(this.redirectOnLogout)
-      window.location.assign(`${ConfigHelper.getAuthContextPath()}signout/${url}`)
+      window.location.assign(`${this.getContextPath()}signout/${url}`)
     } else {
-      window.location.assign(`${ConfigHelper.getAuthContextPath()}signout`)
+      window.location.assign(`${this.getContextPath()}signout`)
     }
   }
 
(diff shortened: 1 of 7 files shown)


vue/sbc-common-components/src/components/SbcHeader.vue
```diff
@@ -231,53 +231,53 @@ export default class SbcHeader extends NavigationMixin {
...
+      window.location.assign(`${this.getContextPath()}signin/bcsc/${url}`)
     } else {
-      window.location.assign(`${ConfigHelper.getAuthContextPath()}signin/bcsc`)
+      window.location.assign(`${this.getContextPath()}signin/bcsc`)
     }
   }
 
   private goToHome () {
     if (this.inAuth) {
-      this.navigateTo(ConfigHelper.getAuthContextPath(), '/home')
```
pksumesh: Except signin and signout all other routes should be pointing to auth

nitheesh-aot: done

vue/sbc-common-components/src/components/SbcHeader.vue
```diff
@@ -231,33 +231,33 @@ export default class SbcHeader extends NavigationMixin {
...
 
   login () {
     if (this.redirectOnLoginSuccess) {
       let url = encodeURIComponent(this.redirectOnLoginSuccess)
       url += this.redirectOnLoginFail ? `/${encodeURIComponent(this.redirectOnLoginFail)}` : ''
-      window.location.assign(`${ConfigHelper.getAuthContextPath()}signin/bcsc/${url}`)
+      window.location.assign(`${this.getContextPath()}signin/bcsc/${url}`)
     } else {
-      window.location.assign(`${ConfigHelper.getAuthContextPath()}signin/bcsc`)
+      window.location.assign(`${this.getContextPath()}signin/bcsc`)
```
sutherlanda: It's not clear to me what this change is doing?  Also, why did these URLs need to be changed?

vue/sbc-common-components/src/components/SbcHeader.vue
```diff
@@ -296,6 +296,12 @@ export default class SbcHeader extends NavigationMixin {
       window.location.assign(`${ConfigHelper.getAuthContextPath()}home`)
     }
   }
+
+  getContextPath (): string {
```
sutherlanda: Does this assume that the client application will have a signIn and signOut view like auth?

sutherlanda:
@nitheesh-aot we attempted this approach already with the navigation-mixin class.  I think we should discuss the approach here, as there were a number of issues with this approach that may be being repeated here.

sutherlanda:
> @nitheesh-aot we attempted this approach already with the navigation-mixin class. I think we should discuss the approach here, as there were a number of issues with this approach that may be being repeated here.

Disregard, @pksumesh has caught me up.

vue/sbc-common-components/src/components/SbcHeader.vue
nitheesh-aot: This is done for getting the context path according to the parent application using our components. so that the signin and signout will be loosely coupled

vue/sbc-common-components/src/components/SbcHeader.vue
nitheesh-aot: Yes, they should have a signin and signout view, which we are already mentioned in the readme doc

PR #126, opened 2020-03-06
removed unnecessary event emit and staff doesnt need to access the account info

saravanpa-aot:
Might fix https://github.com/bcgov/entity/issues/2919

idea is to remove the unnecessary propagation of the events to the sub app. Now on all header refreshes , an event is emitted to the webapp which is not necessary.Only in account switch , the web app has to be notified.

Havent fully verified it.But hoping this will work.Or else will revert/fix it.

### DIFF
--- a/vue/sbc-common-components/src/components/SbcHeader.vue
+++ b/vue/sbc-common-components/src/components/SbcHeader.vue
@@ -196,7 +196,15 @@ export default class SbcHeader extends NavigationMixin {
     if (ConfigHelper.getFromSession(SessionStorageKeys.KeyCloakToken)) {
       const lastUsedAccount = this.getLastAccountId()
       await this.syncUserSettings(lastUsedAccount)
-      this.persistAndEmitAccountId()
+
+      /*
+     Emit event to downstream only when account id is changed.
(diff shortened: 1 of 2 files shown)


vue/sbc-common-components/src/components/SbcHeader.vue
```diff
@@ -193,10 +193,18 @@ export default class SbcHeader extends NavigationMixin {
...
+    if (ConfigHelper.getFromSession(SessionStorageKeys.KeyCloakToken) && (this.accountType !== 'IDIR')) {
       const lastUsedAccount = this.getLastAccountId()
       await this.syncUserSettings(lastUsedAccount)
-      this.persistAndEmitAccountId()
+
+      /*
+     Emit event to downstream only when account id is changed.
+     ie ; on account switch or on when the app is loaded. Otherwise need not to propagate the changes
+     */
+      const isAccountChanged = lastUsedAccount !== this.currentAccount.id.toString()
```
nitheesh-aot: suggetion: better to add a null check for the currentAccount

saravanpa-aot: sure.Will add it

PR #129, opened 2020-03-09
added few null checks

saravanpa-aot:
Few null checks are added to support pending users/staff who might not have an account


vue/sbc-common-components/src/store/modules/account.ts
```diff
@@ -65,9 +65,13 @@ export default class AccountModule extends VuexModule {
 
   @Action({ rawError: true, commit: 'setPendingApprovalCount' })
   public async fetchPendingApprovalCount (): Promise<number> {
-    const response = await AccountService.getPendingMemberCount(this.context.rootState.account &&
-      this.context.rootState.account.currentAccount.id)
-    return (response && response.data && response.data.count) || 0
+    if (this.context.rootState.account && this.context.rootState.account.currentAccount && this.context.rootState.account.currentAccount.id) {
```
sutherlanda: This new check is already 2/3 done in the call to `getPendingMemberCount` below.  Perhaps we can just add the additional check there?

saravanpa-aot: done.

PR #135, opened 2020-03-20
Fix login prompt if token is invalid and refreshed

sutherlanda:
Patch 2.1.19

* Adjust token.services so that if the refresh token is not valid, that KeyCloak will not redirect to the generic login page.  Instead the session will be cleared out.

### DIFF
--- a/vue/sbc-common-components/src/services/token.services.ts
+++ b/vue/sbc-common-components/src/services/token.services.ts
@@ -1,12 +1,16 @@
 import Keycloak, { KeycloakInitOptions, KeycloakInstance } from 'keycloak-js'
 import ConfigHelper from '../util/config-helper'
 import { SessionStorageKeys } from '../util/constants'
+import { getModule } from 'vuex-module-decorators'
+import AuthModule from '../store/modules/auth'
+import { Store } from 'vuex'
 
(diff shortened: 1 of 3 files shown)


vue/sbc-common-components/src/services/token.services.ts
```diff
@@ -129,6 +132,20 @@ class TokenServices {
     }
   }
 
+  private async clearSession () {
+    if (this.store) {
```
saravanpa-aot: May be it needs for fixing the bug ;but just checking ; do we have to invoke the store and set the values ? clearing the value directly from session-storage wont work? If session storage clearing works , i think we wont need the store here and this component can stay disconnected with the store.
I am thinking how the behaviour will be when the user is in coops. Can the same bug happen when user stays in coops?

sutherlanda: We should clear both the store and session storage.  Ideally we shouldn't be putting anything in session storage at all, but I left it in so as not to require a change fro co-ops.

The root cause was just the 'login-required', not stuff hanging around in session storage.  Either way, if are sure to clear the store and session storage on logout, this will eliminate any chance of stale data being used on the next login.

PR #137, opened 2020-03-26
conditionally avoided notification count call for bcros users

nitheesh-aot:
conditionally avoided notification count call for bcros users


vue/sbc-common-components/src/store/modules/account.ts
```diff
@@ -72,7 +72,9 @@ export default class AccountModule extends VuexModule {
     if (response && response.data) {
       const orgs = response.data.filter(userSettings => (userSettings.type === 'ACCOUNT'))
       this.context.commit('setCurrentAccount', currentAccountId ? orgs.find(org => String(org.id) === currentAccountId) : orgs[0])
-      await this.context.dispatch('fetchPendingApprovalCount')
+      if (this.currentUser?.loginSource !== 'BCROS') {
```
sutherlanda: It may be better to just check this if the loginSource === 'BCSC'.  That way we avoid checking for IDIR users as well.

nitheesh-aot: make sense, will change that
thanks @sutherlanda

PR #140, opened 2020-03-31
3007 product selector

sutherlanda:
@jeznorth feel free to add to this PR if needed.


sutherlanda:
> Just want to confirm, the product selectors will be same regardless if the user is authenticated and different account have diff products?

I think for now yeah.  'MVP' for this feature will just show all products until we have the subscription part of the account settings sorted out.  Once we have that in place, we would update this feature to only show the products that account has subscribed to.

PR #141, opened 2020-04-09
Add missing promise resolve/reject

sutherlanda:
Patch 2.1.25: Fix a bug with token refresh where a promise will never resolve if not authenticated.

### DIFF
--- a/vue/sbc-common-components/package.json
+++ b/vue/sbc-common-components/package.json
@@ -1,6 +1,6 @@
 {
   "name": "sbc-common-components",
-  "version": "2.1.24",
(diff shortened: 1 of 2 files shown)


sutherlanda:
> Changes looks good..
> Just being on the side of caution ,should we let others know about this new behaviour? do you think they will have to add the same behaviour that we added in sbc-auth?

The change in common is pretty much just correcting broken behaviour.  The changes in auth are not necessarily applicable to the other apps.

byshine:
Thanks for this @sutherlanda This will allow our app (director search) to catch the error.

PR #149, opened 2020-05-13
Configure sbc-common-components LaunchDarkly for feature flags#2439

nitheesh-aot:
Configure sbc-common-components LaunchDarkly for feature flags#2439

### DIFF
--- a/vue/sbc-common-components/src/components/SbcHeader.vue
+++ b/vue/sbc-common-components/src/components/SbcHeader.vue
@@ -143,6 +143,7 @@ import AccountModule from '../store/modules/account'
 import AuthModule from '../store/modules/auth'
 import { KCUserProfile } from '../models/KCUserProfile'
 import keycloakService from '../services/keycloak.services'
+import LaunchDarklyService from '../services/launchdarkly.services'
 import SbcProductSelector from './SbcProductSelector.vue'
 
 declare module 'vuex' {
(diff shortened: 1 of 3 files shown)


saravanpa-aot:
@severinbeauvais Sev, do u have time to look at this?

vue/sbc-common-components/src/components/SbcHeader.vue
```diff
@@ -205,20 +214,17 @@ export default class SbcHeader extends Mixins(NavigationMixin) {
 
   get showAccountSwitching (): boolean {
     try {
-      const flags = JSON.parse(ConfigHelper.getFromSession(SessionStorageKeys.LaunchDarklyFlags) || '{}')
-      return flags && flags['account-switching']
+      return LaunchDarklyService.getFlag('account-switching')
```
severinbeauvais, line 217: Why is this code (inside try..catch) different from code below (line 224), when they do the same thing?
If you're looking for a suggestion which way to go, I think the code below looks cleaner (if in fact it works as intended).

vue/sbc-common-components/src/services/launchdarkly.services.ts
```diff
@@ -0,0 +1,61 @@
+import { initialize, LDFlagSet } from 'launchdarkly-js-client-sdk'
```
severinbeauvais, line 1: This looks OK, but I'm wondering why it doesn't look more like this https://github.com/bcgov/business-filings-ui/blob/master/src/common/FeatureFlags.ts ?

Would you recommend that Entities team update their implementation to more closely match yours?

vue/sbc-common-components/src/components/SbcHeader.vue
nitheesh-aot: It was already inside the try catch, but yeah, the below code does the same and works,
I will change this to make it consistant
Thanks @severinbeauvais

vue/sbc-common-components/src/services/launchdarkly.services.ts
nitheesh-aot: It wont be necessary, I have done it this way since we will be using the session storage to save the flags. so the initialization and all done inside the class itself, only if the flags are not available in the session. which is for avoiding the unnecessary calls to LD.

PR #150, opened 2020-05-15
removed readonly province and country from BaseAddress component

severinbeauvais:
*Issue #:* /bcgov/entity#3477

*Description of changes:*
- removed confusing readonly rules
- updated unit tests accordingly
- update version to 2.1.29
- committed package-lock.json <<< this is a best practice so that server builds === local builds

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the business-filings-ui license (Apache 2.0).

### DIFF
--- a/vue/sbc-common-components/src/components/BaseAddress.vue
+++ b/vue/sbc-common-components/src/components/BaseAddress.vue
@@ -72,7 +72,6 @@
                     v-model="addressLocal.addressRegion"
                     :items="getCountryRegions(addressCountry)"
                     :rules="[...rules.addressRegion, ...spaceRules]"
-                    :readonly="isSchemaBC()"
           />
           <v-text-field v-else
                         filled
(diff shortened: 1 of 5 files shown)


vue/sbc-common-components/src/components/BaseAddress.vue
```diff
@@ -72,15 +72,13 @@
                     v-model="addressLocal.addressRegion"
                     :items="getCountryRegions(addressCountry)"
                     :rules="[...rules.addressRegion, ...spaceRules]"
-                    :readonly="isSchemaBC()"
```
severinbeauvais, line 75: These rules seemed like a good idea at the time, but some users would not use AddressComplete to find an address (or their address was not in AddressComplete). Because these components were readonly, users were not able to enter a province or country and could therefore not enter a valid address, which blocked some of them from completing their form.

PR #160, opened 2020-06-24
Fix for https://github.com/bcgov/entity/issues/4133

saravanpa-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/4133

*Description of changes:*
in the sbc-header , added the logic for fetching name from DB if the user is bceid user
in the switch cases for accounts ,added bceid as well

### DIFF
--- a/vue/sbc-common-components/src/components/SbcHeader.vue
+++ b/vue/sbc-common-components/src/components/SbcHeader.vue
@@ -100,7 +100,7 @@
               </v-list-item-content>
             </v-list-item>
             <!-- BEGIN: Hide if authentication is IDIR -->
-            <v-list-item @click="goToUserProfile()" v-if="isBCSC">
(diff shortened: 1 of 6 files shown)


vue/sbc-common-components/src/components/SbcHeader.vue
```diff
@@ -264,17 +265,35 @@ export default class SbcHeader extends Mixins(NavigationMixin) {
...
+
+  @Watch('isAuthenticated')
+  private async onisAuthenticated (isAuthenitcated: string, oldVal: string) {
+    if (isAuthenitcated) {
+      await this.updateProfile()
+    }
+  }
+
+  private async updateProfile () {
+    if (this.isBceid) {
```
sumesh-aot, line 295: Is this correct? isBceid is a function right ?

saravanpa-aot: a getter..so will work i guess

PR #165, opened 2020-07-06
[entity4220] UI Updates for Authentication Options

jeznorth:
Updating the product selector

- Using a simple list until we have a more products available in BC Registries
- Code clean-up / refactor (accessibility improvements, mobile improvements)

### DIFF
--- a/vue/sbc-common-components/src/components/SbcAuthenticationOptions.vue
+++ b/vue/sbc-common-components/src/components/SbcAuthenticationOptions.vue
@@ -1,42 +1,48 @@
 <template>
   <v-container class="view-container">
-    <div class="view-header flex-column mb-10">
+    <div class="view-header flex-column mb-9">
       <h1 class="view-header__title">Log in to BC Registries</h1>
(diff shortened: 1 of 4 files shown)


vue/sbc-common-components/src/components/SbcAuthenticationOptions.vue
```diff
@@ -1,42 +1,48 @@
...
         v-for="authOption in authOptions"
         :key="authOption.type"
       >
         <v-card
           flat
-          class="account-card text-center pa-7 d-flex flex-column"
+          outlined
+          hover
+          class="account-card text-center pa-10 elevation-2 d-flex"
+          @click="selectAuthType(authOption)"
```
nitheesh-aot, line 22: Is this click is needed here? as per the design, we have the click on the button in each card

jeznorth: I've update this to more closely match the patterns we've established elsewhere in the application (full-clickable cards)

PR #168, opened 2020-07-10
BCSC - Frozen on one screen#4300

nitheesh-aot:
director search application needs a way to show terms of service to users bcgov/entity#4271
BCSC - Frozen on one screen bcgov/entity#4300

### DIFF
--- a/vue/sbc-common-components/src/components/SbcSignin.vue
+++ b/vue/sbc-common-components/src/components/SbcSignin.vue
@@ -34,7 +34,11 @@ import NavigationMixin from '../mixins/navigation-mixin'
     }
     this.$options.methods = {
       ...(this.$options.methods || {}),
-      ...mapActions('account', ['loadUserInfo', 'syncAccount'])
+      ...mapActions('account', [
+        'loadUserInfo',
+        'syncAccount',
+        'getCurrentUserProfile'
+      ])
     }
   }
 })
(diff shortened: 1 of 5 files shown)


vue/sbc-common-components/src/components/SbcSignin.vue
```diff
@@ -62,24 +62,8 @@ export default class SbcSignin extends NavigationMixin {
...
-                break
-            }
-          }
-
           // emit an event to parent that user profile syncing is completed.
-          // the argument isRedirectToCreateAccount will be used to check in the parent signin view component
-          // to determine whether to perform the redirect on that end,
-          // if the value is true, then the redirection should be happening from this component.
-          this.$emit('sync-user-profile-ready', isRedirectToCreateAccount)
+          this.$emit('sync-user-profile-ready')
```
sumesh-aot, line 66: What is the change here ? If the user is not on accounts application, how is this going to work ? I think this change breaks the whole redirection you did as part of previous ticket to redirect if the user is not part of any account. Please correct if I am wrong.

nitheesh-aot: Yeah, but actually we are handling the redirection in our auth app.
User will come to the auth app anyways for the signin process.
and the redirectUrl prop from them is actually making the redirection back to their app.
I have discussed with saravan and intercepted that part and made the changes here. which will be much better maintainable than this.

https://github.com/bcgov/sbc-auth/pull/870

this PR is having the changes

PR #169, opened 2020-07-14
redirection only for non sbc-auth routes

nitheesh-aot:
- redirection logics is now checking only for non sbc-auth routes
- version update to 2.2.3

### DIFF
--- a/vue/sbc-common-components/src/components/SbcSignin.vue
+++ b/vue/sbc-common-components/src/components/SbcSignin.vue
@@ -67,23 +67,26 @@ export default class SbcSignin extends NavigationMixin {
           // sync the account if there is one
           await this.syncAccount()
 
-          const currentUser = await this.getCurrentUserProfile()
-
           // redirect to create account page if the user has no 'account holder' role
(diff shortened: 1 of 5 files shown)


vue/sbc-common-components/src/components/SbcSignin.vue
```diff
@@ -67,23 +67,26 @@ export default class SbcSignin extends NavigationMixin {
           // sync the account if there is one
           await this.syncAccount()
 
-          const currentUser = await this.getCurrentUserProfile()
-
           // redirect to create account page if the user has no 'account holder' role
           const isRedirectToCreateAccount = (userInfo.roles.includes(Role.PublicUser) && !userInfo.roles.includes(Role.AccountHolder))
```
saravanpa-aot, line 71: there is some logic in auth-web related to isRedirectToCreateAccount rite?Do we stil need that?

nitheesh-aot: wont need that, I will be removing that

saravanpa-aot: cool..

PR #182, opened 2020-08-17
Trim trailing slash in api and context urls

nitheesh-aot:
- Trim trailing slash in api and context urls


nitheesh-aot:
> Changes look fine.
> Can this be linked to a ticket?
> Also, how tested is this?

I have done the dev testing, it will remove the trailing slash only if its available, otherwise it will use that url as it is

PR #185, opened 2020-08-28
4780: Update user profile only during login

sumesh-aot:
v2.3.4

Issue:
User Profile is not updated during login, which makes it hard to audit user session start time.
https://github.com/bcgov/entity/issues/4780

Fix:
Added updateUserProfile on successful login.

### DIFF
--- a/vue/sbc-common-components/src/components/SbcSignin.vue
+++ b/vue/sbc-common-components/src/components/SbcSignin.vue
@@ -37,7 +37,8 @@ import NavigationMixin from '../mixins/navigation-mixin'
       ...mapActions('account', [
         'loadUserInfo',
         'syncAccount',
-        'getCurrentUserProfile'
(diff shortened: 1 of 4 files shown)


vue/sbc-common-components/src/components/SbcSignin.vue
```diff
@@ -64,6 +66,10 @@ export default class SbcSignin extends NavigationMixin {
           await KeyCloakService.initSession()
           // tell KeycloakServices to load the user info
           const userInfo = await this.loadUserInfo()
+
+          //update user profile
+          await this.updateUserProfile()
```
saravanpa-aot, line 71: hope you tested well and made sure this is called only during the signin ...

sumesh-aot: Yea, did some testing.

PR #194, opened 2020-10-27
updated BaseAddress to latest AddressComplete + fixes

severinbeauvais:
*Issue #:* /bcgov/entity#988 /bcgov/entity#2551

*Description of changes:*
- fixed BaseAddress to use latest AddressComplete
- fixed BaseAddress to support multiple instances of itself
- fixed BaseAddress to support tabbing to search field (ie, keypress event) (# 2551)
- locally cached AddressComplete v2.30 files
- updated version to 2.4.2

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the bcrs-entities-create-ui license (Apache 2.0).


vue/sbc-common-components/package.json
```diff
@@ -25,6 +25,7 @@
     "country-list": "^2.2.0",
     "keycloak-js": "^9.0.3",
     "launchdarkly-js-client-sdk": "^2.16.1",
+    "lodash.uniqueid": "^4.0.1",
```
severinbeauvais, line 28: This is kind of silly because the entire lodash library is a dependency of other packages we depend on, but I'm being explicit :)

vue/sbc-common-components/public/index.html
```diff
@@ -5,13 +5,18 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
     <link rel="icon" href="<%= BASE_URL %>favicon.ico">
+    <link rel="stylesheet" href="<%= BASE_URL %>css/addresscomplete-2.30.css">
+    <script type="text/javascript" src="<%= BASE_URL %>js/addresscomplete-2.30.js" defer></script>
```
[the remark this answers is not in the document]
severinbeauvais: PS I copied/cached the css and js files locally so we never hang on startup if the Canada Post website is slow/down.

vue/sbc-common-components/src/components/BaseAddress.vue
```diff
@@ -34,15 +34,18 @@
...
-          <!-- NB: "name" attribute is needed for moveElementId() -->
-          <v-text-field autocomplete="address-complete"
+          <!-- NB1: AddressComplete needs to be enabled each time user clicks in this search field.
+               NB2: Only process first keypress -- assumes if user moves between instances of this
+                   component then they are using the mouse (and thus, clicking). -->
+          <v-text-field autocomplete="off"
                         filled
                         class="street-address"
-                        name="street-address"
+                        :id="streetAddressId"
```
severinbeauvais, line 43: Using a unique ID here (and for country) does away with the old "search for name and move the id" code that didn't support multiple instances of this component anyway. And it's simpler!

vue/sbc-common-components/src/components/BaseAddress.vue
```diff
@@ -255,25 +264,30 @@ export default class BaseAddress extends Mixins(ValidationMixin, CountriesProvin
...
   /**
-   * Emits the Vuelidate state of the address entered by the user.
+   * Watches changes to the Schema object, so that if the parent changes the data, then
+   * the working copy of it is updated.
    */
-  @Emit('valid')
-  private emitValid (val: boolean): void { }
+  @Watch('schema', { deep: true, immediate: true })
+  private onSchemaChanged (): void {
+    this.schemaLocal = { ...this.schema }
```
severinbeauvais, line 289: I made this a watcher because the previous code was giving me the "using a property before it's declared" error. So it's set immediately which does the same thing but without the error.

vue/sbc-common-components/src/components/BaseAddress.vue
```diff
@@ -316,76 +330,48 @@ export default class BaseAddress extends Mixins(ValidationMixin, CountriesProvin
...
 
-  /**
-   * Sets the id attribute of the _named_ element to its name. If there was a pre-existing element with the id already
-   * set, it will be unset.
-   *
-   * @param name the name of the element for which to set the id.
-   */
-  private moveElementId (name: string): void {
-    const oldElement = document.getElementById(name)
-    const thisElement = this.$el.querySelector('[name="' + name + '"]')
```
severinbeauvais, line 360: This only worked for the first instance of the BaseAddress component on a page, but there are many cases where there are multiple addresses on the same page.

vue/sbc-common-components/src/components/BaseAddress.vue
```diff
@@ -316,76 +330,48 @@ export default class BaseAddress extends Mixins(ValidationMixin, CountriesProvin
...
     // Set up the two fields that AddressComplete will use for input.
     // Ref: https://www.canadapost.ca/pca/support/guides/advanced
     // Note: Use special field for country, which user can't click, and which AC will overwrite
     //       but that we don't care about.
-    //       (Option `populate: false` doesn't seem to work.)
     const fields = [
-      { element: 'street-address', mode: pca.fieldMode.SEARCH },
-      { element: 'address-country-pca', mode: pca.fieldMode.COUNTRY }
+      { element: this.streetAddressId, field: 'Line1', mode: pca.fieldMode.SEARCH },
+      { element: this.addressCountryId, field: 'CountryName', mode: pca.fieldMode.COUNTRY }
```
severinbeauvais, line 374: I added "field" to work with latest AddressComplete library.

Also this is where we tell AddressComplete about the elements to use (using unique ids!).

vue/sbc-common-components/public/index.html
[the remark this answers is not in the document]
cameron-freshworks: Pretty cool! 
Does caching like this have any impact on load times etc. 
I imagine in this case, the content is probably relatively small but for examples sake, would doing something like this with a large library have an impact?

vue/sbc-common-components/src/components/BaseAddress.vue
```diff
@@ -401,13 +387,12 @@ export default class BaseAddress extends Mixins(ValidationMixin, CountriesProvin
...
    * Callback to update the address data after the user chooses a suggested address.
-   *
-   * @param address the data object returned by the AddressComplete Retrieve API.
+   * @param address the data object returned by the AddressComplete Retrieve API
    */
   private addressCompletePopulate (address: object): void {
     const newAddressLocal: object = {}
 
-    newAddressLocal['streetAddress'] = address['Line1']
+    newAddressLocal['streetAddress'] = address['Line1'] || 'N/A'
```
severinbeauvais, line 395: While testing, I noticed some address (eg, in Algeria) don't have any LineX data, so I added this fallback so that an address filled in my AddressComplete is valid (since street address is typically a required field).

vue/sbc-common-components/src/components/BaseAddress.vue
```diff
@@ -160,52 +164,54 @@ export default class BaseAddress extends Mixins(ValidationMixin, CountriesProvin
...
 
-  /**
-   * Getter for Address Country, to simplify template and so we can watch it below.
-   */
+  /** A unique id for this instance of this component. */
+  private uniqueId = uniqueId()
+
+  /** A unique id for the Street Address input. */
+  private get streetAddressId (): string {
+    return `street-address-${this.uniqueId}`
```
cameron-freshworks, line 186: Off the top of your head, what is the difference between lodash's uniqueId and something like UUID?

vue/sbc-common-components/public/index.html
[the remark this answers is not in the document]
severinbeauvais: Of course it has an impact, but I assume it's faster to load from our server than an external server (esp if the external server is down).
Note also that I use the "defer" prop to load the script as late as possible in the app loading flow.

cameron-freshworks (approved):
Looks good! Just had a few questions.

vue/sbc-common-components/src/components/BaseAddress.vue
severinbeauvais: This lodash method returns a simple id starting from '1' and incrementing each time you call it.

The UUID package returns a random uuid string, which would have been total overkill here.

cameron-freshworks: Yea, the UUID string is pretty intense. Thanks for the info!

severinbeauvais:
If the next reviewer feels qualified, please merge this.

vue/sbc-common-components/public/index.html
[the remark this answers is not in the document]
saravanpa-aot: if render blocking is the issue , should async wont work?
https://developers.google.com/speed/docs/insights/BlockingJS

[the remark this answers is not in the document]
severinbeauvais: It might work but it's deprecated 😛

saravanpa-aot (commented):
I put the comment under a resolved review..so putting it here again..
isnt the async loading better than caching locally?
I am hoping async attribute will help if renderblocking js is the problem
https://developers.google.com/speed/docs/insights/BlockingJS

vue/sbc-common-components/public/index.html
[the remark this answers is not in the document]
saravanpa-aot: what is deprecated? async attirbute?

severinbeauvais:
> I put the comment under a resolved review..so putting it here again..
> isnt the async loading better than caching locally?
> I am hoping async attribute will help if renderblocking js is the problem
> https://developers.google.com/speed/docs/insights/BlockingJS

Caching locally is to prevent blocking if the Canada Post server is slow/down.

I can't use "async" because it's deprecated/shut down, as per the link your posted.

vue/sbc-common-components/public/index.html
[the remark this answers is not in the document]
severinbeauvais: I just looked again. Ooops, API is deprecated not "async".

I believe "defer" provides faster initial loading than "async" here.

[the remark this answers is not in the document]
saravanpa-aot: hm.. did it still block the rendering [slowed the page loading] , when used async?

vue/sbc-common-components/package.json
saravanpa-aot: he he..i am good with the change..but generally moving away from loadash is advised..  es6 alternatives are there for most of the cases..

vue/sbc-common-components/public/index.html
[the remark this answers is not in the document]
saravanpa-aot: but Thanks for taking this up and pushing this change..I am just curious why async is still blocking the rendering

vue/sbc-common-components/package.json
severinbeauvais: I agree with you.

In this case there are other alternatives, but this was the easy/quick/clean way :)

vue/sbc-common-components/public/index.html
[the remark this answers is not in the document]
severinbeauvais: I don't remember my test results when I initially tried async vs defer and decided on the latter.

(Would async load scripts at the same time? If so then this would reduce the bandwidth available for other loads.)

(Also I think defer doesn't autorun the script until everything else is loaded, which is desired here -- there is no need to run the AddressComplete init method until after the app is loaded.)

severinbeauvais:
Saravan has agreed that I can merge this PR.

PR #198, opened 2020-11-05
Authentication Options Dialog

cameron-freshworks:
*Issue #:* /bcgov/entity#5332

*Description of changes:*

* Implemented Dialog version of SbcAuthenticationOptions. 

* Added README for the new component. 

* Invision design: https://projects.invisionapp.com/d/main#/console/20760950/436781801/preview

Screenshot if permissions are blocking:
Screen Shot 2020-11-05 at 10 26 37 AM

### DIFF
--- a/vue/sbc-common-components/src/components/SbcAuthenticationOptionsDialog.vue
+++ b/vue/sbc-common-components/src/components/SbcAuthenticationOptionsDialog.vue
@@ -1,0 +1,160 @@
+<template>
+  <v-dialog v-model="showModal" max-width="850px" persistent>
+    <v-card class="pa-4">
+      <v-container class="view-container">
+        <div class="flex-column mb-9">
+          <v-row>
+            <v-col cols="11">
+              <h3>Log in to BC Registries</h3>
+            </v-col>
+            <v-col cols="1">
+              <v-icon large color="primary" @click="emitClose()">mdi-close</v-icon>
+            </v-col>
+          </v-row>
+          <p>
+            Don't have a BC Registries account?
+            <a @click="goToCreateAccount">
+              <u>Create an account</u>
+            </a>
+          </p>
+        </div>
+        <v-row>
+          <v-col
+            sm="12"
+            md="6"
+            class="d-flex align-stretch"
+            v-for="authOption in authOptions"
+            :key="authOption.type"
+          >
+            <v-card
+              flat
+              outlined
+              hover
+              class="account-card text-center pa-10 elevation-2 d-flex"
+              @click="selectAuthType(authOption)"
+              :ripple="false"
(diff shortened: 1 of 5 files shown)


cameron-freshworks:
Just adding a screenshot from it's implementation in NameRequest for some context.

nitheesh-aot:
> Just adding a screenshot from it's implementation in NameRequest for some context.

Can you add a screenshot instead of the design link? link is asking permission

cameron-freshworks:
> > Just adding a screenshot from it's implementation in NameRequest for some context.
> 
> Can you add a screenshot instead of the design link? link is asking permission

Sorry, just did this above. I forgot to hit 'send' 😁

vue/sbc-common-components/src/components/SbcAuthenticationOptionsDialog.vue
```diff
@@ -0,0 +1,160 @@
+<template>
+  <v-dialog v-model="showModal" max-width="850px" persistent>
+    <v-card class="pa-4">
+      <v-container class="view-container">
```
nitheesh-aot, line 4: There is a SbcAuthenticationOptions.vue component which has the same design and view of these. Can we leverage that to this, rather than repeating the code here?

cameron-freshworks: Hmm yea, it is very similar. After discussing with Jeremy we thought perhaps it would be better to create a new component specific to this scenario (dialog) instead of refactoring/complicating the existing full view component. 
Would you prefer to refactor the existing one?

nitheesh-aot: nope, refactoring wont be a good idea.
but we can include 'SbcAuthenticationOptions' as a child component here.
and include in the html inside the dialog container like <SbcAuthenticationOptions></SbcAuthenticationOptions >. and extend those functionalities to this modal popup.

cameron-freshworks: Ok, a sub template? With a prop determining which template to use?

cameron-freshworks: It might make sense to make the dialog a sub template of the full view, no?

nitheesh-aot: no, the other way will be better.

SbcAuthenticationOptions as a sub template in this Dialog popup component.

we can define the same prop names in the dialog component to pass the values between the components

nitheesh-aot (approved):

PR #201, opened 2020-12-29
User in-app notification of suspended account

shabeeb-aot:
Issue #: bcgov/entity#5229

Description of changes:
* If the account is suspended status, the user will redirect to the account-freeze page according to role
* updated version 2.4.6 to 2.4.7

### DIFF
--- a/vue/sbc-common-components/src/components/SbcHeader.vue
+++ b/vue/sbc-common-components/src/components/SbcHeader.vue
@@ -408,6 +408,8 @@ export default class SbcHeader extends Mixins(NavigationMixin) {
       await this.loadUserInfo()
       await this.syncAccount()
       await this.updateProfile()
+      // checking for account status
+      await this.checkAccountStatus()
     }
   }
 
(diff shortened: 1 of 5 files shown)


vue/sbc-common-components/src/components/SbcHeader.vue
```diff
@@ -462,6 +464,13 @@ export default class SbcHeader extends Mixins(NavigationMixin) {
     this.redirectToPath(this.inAuth, `${Pages.ACCOUNT}/${this.currentAccount.id}/${Pages.SETTINGS}/transactions`)
   }
 
+  private checkAccountStatus () {
+    // redirect if accoutn status is suspended
+    if (this.currentAccount?.accountStatus && this.currentAccount?.accountStatus === 'NSF_SUSPENDED') {
```
sumesh-aot, line 469: May be get this status from an array? I think there will be other suspended status as well like SUSPENDED.

sumesh-aot (approved):

saravanpa-aot (approved):

vue/sbc-common-components/src/components/SbcHeader.vue
shabeeb-aot: will update in the next PR. Now we have only one value, so its good to make a direct check

PR #207, opened 2021-03-10
Clean Up Tasks to Support Government Ministry Accounts

shabeeb-aot:
* remove IDIR check in header component

### DIFF
--- a/vue/sbc-common-components/src/util/constants.ts
+++ b/vue/sbc-common-components/src/util/constants.ts
@@ -33,7 +33,7 @@ export enum LoginSource {
 export enum Role {
   AccountHolder = 'account_holder',
   PublicUser = 'public_user',
-  STAFF = 'staff',
+  Staff = 'staff',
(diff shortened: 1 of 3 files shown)


karthik-aot (approved):

saravanpa-aot (approved):
Good for now. But next sprint , I think we will remove to work on create account button and multiple account for govm users.

vue/sbc-common-components/src/util/constants.ts
```diff
@@ -32,7 +32,9 @@ export enum LoginSource {
 
 export enum Role {
   AccountHolder = 'account_holder',
-  PublicUser = 'public_user'
+  PublicUser = 'public_user',
+  STAFF = 'staff',
```
sumesh-aot, line 36: Can you rename STAFF as Staff to keep consistent with other enum keys in Role ?

shabeeb-aot: updated

PR #209, opened 2021-03-16
Add  browser version warning message component.

pwei1018:
https://github.com/bcgov/entity/issues/6764 

Add BrowserVersionAlert vue component;
If Browser version is IE 11 or under, the alert message will show up;

image

### DIFF
--- a/vue/sbc-common-components/src/components/BrowserVersionAlert.vue
+++ b/vue/sbc-common-components/src/components/BrowserVersionAlert.vue
@@ -1,0 +1,94 @@
+<template>
+  <sbc-system-banner
+    :show="browserSupported === false"
+    type="warning"
+    :message="alertMessage"
+  ></sbc-system-banner>
+</template>
+
+<script lang='ts'>
+import { Component, Prop, Vue } from 'vue-property-decorator'
+import SbcSystemBanner from './SbcSystemBanner.vue'
+
+@Component({
+  components: {
+    SbcSystemBanner
+  }
+})
+export default class BrowserVersionAlert extends Vue {
+  @Prop({ default: `This website will not work properly in your current browser. We recommend viewing this website
+  in a newer browser. BC Registries and Online Services websites work better when viewed in the latest versions of
+  most popular new browsers.` })
+  readonly alertMessage: string
+
+  @Prop({ default: true })
+  browserSupported: boolean
+
+  private getBrowser (): { name: string; version: number } {
+    let ua = navigator.userAgent
+    let tem: string[]
+    let M = ua.match(/(opera|chrome|safari|firefox|msie|trident(?=\/))\/?\s*(\d+)/i) || []
+
+    if (/trident/i.test(M[1])) {
+      tem = /\brv[ :]+(\d+)/g.exec(ua) || []
+      return { name: 'IE', version: parseInt(tem[1] || '') }
+    }
+    if (M[1] === 'Chrome') {
+      tem = ua.match(/\bOPR\/(\d+)/)
+      if (tem != null) { return { name: 'Opera', version: parseInt(tem[1]) } }
+    }
+
+    // MS Edge Html
+    if (window.navigator.userAgent.indexOf('Edge') > -1) {
+      // eslint-disable-next-line no-useless-escape
+      tem = ua.match(/\Edge\/(\d+)/)
+
+      if (tem != null) {
+        return { name: 'Edge', version: parseInt(tem[1]) }
+      }
+    }
+
+    // MS Edge Chromium
+    if (window.navigator.userAgent.indexOf('Edg') > -1) {
+      // eslint-disable-next-line no-useless-escape
+      tem = ua.match(/\Edg\/(\d+)/)
+
+      if (tem != null) {
+        return { name: 'Edge', version: parseInt(tem[1]) }
+      }
+    }
+
+    M = M[2] ? [M[1], M[2]] : [navigator.appName, navigator.appVersion, '-?']
+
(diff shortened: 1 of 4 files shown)


severinbeauvais:
Is the banner message approved by a PO?

vue/sbc-common-components/src/components/BrowserVersionAlert.vue
```diff
@@ -0,0 +1,110 @@
+<template>
+  <sbc-system-banner
+    v-if="browserSupported === false"
+    v-bind:show="browserSupported === false"
```
severinbeauvais, line 4: You don't need both v-if and v-show.
(v-if conditionally renders the component)
(v-show conditionally set the display/none attribute, so the component is there but not visible, which may affect screen readers)
Also, a shortcut to "v-bind:show=" is simply ":show=". (Also line 6 below.)

vue/sbc-common-components/src/components/BrowserVersionAlert.vue
```diff
@@ -0,0 +1,110 @@
...
+      return aPath.every((p) => {
+        m = m._children[p]
+        return m
+      })
+    }
+  }
+})
+export default class BrowserVersionAlert extends Vue {
+  private alertMessage: string = ''
+  private browserSupported: boolean = true
```
severinbeauvais, line 36: Do these props need the `@Prop()` decorator? (The sbc code is a bit different from the Entity UI code so I really don't know if this is needed or helpful here.)

vue/sbc-common-components/src/components/BrowserVersionAlert.vue
```diff
@@ -0,0 +1,110 @@
...
+      M.splice(1, 1, tem[1])
+    }
+    return {
+      name: M[0],
+      version: +M[1]
+    }
+  }
+
+  private isSupported (browser: { name: string; version: number }): boolean {
+    var supported = true
```
severinbeauvais, line 84: Please use `let` or `const` instead of var.

vue/sbc-common-components/src/components/BrowserVersionAlert.vue
```diff
@@ -0,0 +1,110 @@
...
+    }
+    return supported
+  }
+
+  private async mounted () {
+    this.browserSupported = this.isSupported(this.getBrowser())
+    if (!this.browserSupported) {
+      // render unsupported message
+      this.alertMessage = 'Unsupported Browser! We only support the recent versions of major browsers like Chrome, Firefox, Safari, and Edge.'
+      console.log(this.alertMessage)
```
severinbeauvais, line 97: This line may need a "lint disable" comment. (Depends on lint configuration whether console.xxx() generated a warning/error.) Even if not needed, it's probably good practise to add such a comment anyway.

saravanpa-aot (commented):
hm.. if we put this in sbc-header , it might be better since everybody uses it...

severinbeauvais:
> hm.. if we put this in sbc-header , it might be better since everybody uses it...

I agree.

vue/sbc-common-components/src/components/BrowserVersionAlert.vue
pwei1018: Changed

vue/sbc-common-components/src/components/BrowserVersionAlert.vue
```diff
@@ -0,0 +1,110 @@
...
+  interface Store<S> {
+    hasModule(_: string[]): boolean
+  }
+}
+
+@Component({
+  components: {
+    SbcSystemBanner
+  },
+  beforeCreate () {
```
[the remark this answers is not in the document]
pwei1018: Removed.

vue/sbc-common-components/src/components/BrowserVersionAlert.vue
pwei1018: Changed

vue/sbc-common-components/src/components/BrowserVersionAlert.vue
pwei1018: Fixed

vue/sbc-common-components/src/components/BrowserVersionAlert.vue
pwei1018: Removed.

pwei1018 (commented):
@saravanpa-aot @severinbeauvais I put the banner into sbc header component. The only issue is the banner seems display overlap the main content. 

image

vue/sbc-common-components/src/components/SbcHeader.vue
```diff
@@ -287,7 +287,11 @@
         </v-btn>
       </div>
     </v-container>
+    <div id="browse-version-alert">
+      <BrowserVersionAlert />
```
severinbeauvais, line 291: Where is the existing banner code? Would it make sense to put this in the same place?

pwei1018: Not in the header. Header doesn't include the banner.

severinbeauvais: OK.

severinbeauvais:
> @saravanpa-aot @severinbeauvais I put the banner into sbc header component. The only issue is the banner seems display overlap the main content.

The overlap is a (small) problem. Can you please check with @jeznorth on the necessary CSS so it doesn't overlap. Also check with him on the overall look of this warning banner.

Also, was the warning message approved by a PO (and/or UX person)?

severinbeauvais (approved):

PR #220, opened 2021-05-27
Fix bug of empty line in address display

eve-git:


severinbeauvais:
Please update description block at top of PR -- see other PRs for the template text.

severinbeauvais:
Also, this PR refers to ticket 2301, which was actually closed in Jan 2020. Did you mean to refer to ticket 2312 instead?

vue/sbc-common-components/src/components/BaseAddress.vue
```diff
@@ -458,8 +458,8 @@ export default class BaseAddress extends Mixins(ValidationMixin, CountriesProvin
   }
 }
 
-.pre-wrap {
-  white-space: pre-wrap;
+.pre-line {
+  white-space: pre-line;
```
severinbeauvais, line 462: Did you test this in Edge, Chrome, FF? Safari if you're able? (Maybe check IE11 as well though we don't try hard to support it.)

severinbeauvais (changes requested):
According to comments in ticket 2301, this has already been fixed. Therefore, what does this code change actually do? Can it be demonstrated that the issue reproduces before the fix and does not reproduce after the fix?

vue/sbc-common-components/src/components/BaseAddress.vue
eve-git: I tested in Edge, IE, Chrome and FF.

severinbeauvais: Was it broken before, and fixed now?

severinbeauvais: Also, what do the addresses long like with long lines?

You can paste screen shots here to show your testing (or paste in ticket).

eve-git: in FF, before change:
image

eve-git: after code change in FF:
image

severinbeauvais: OK, the fix above looks good. But it seems to undo a previous fix, so I'm wondering if this actually breaks something else? Like long address lines maybe?

eve-git: What does long address lines mean? The max length of the address line is 50 and it is good -- is it sufficient?

severinbeauvais: In the example above, it looks like 50 chars will easily fit on the line, but in a different example (or possibly on a narrower screen), 50 chars might not fit on the line. In that case, the address needs to wrap correctly.

I just had a look at https://www.w3schools.com/cssref/pr_text_white-space.asp ... Both the previous and new CSS properties appear to support wrapping, so you're probably OK, but please verify it.

As a general concept, since there are no unit tests here to verify whitespace behaviour, you will have to manually verify the different scenarios. I can think of the following (there may  be more):
- does long address line wrap correctly?
- is whitespace collapsed or preserved?

eve-git: Comparing pre-wrap and pre-line, they all wrap text to fit the container. pre-wrap does NOT collapse whitespace while pre-line collapse extra whitespace to one. To my understanding, replace pre-wrap with pre-line should fix the bug and no other issues related.

shabeeb-aot (approved):

severinbeauvais:
I believe another update went into the sbc-common-components (and is published as 2.4.16). Therefore, please update the version in this PR to 2.4.17.

eve-git:
I renamed the branch to 2312 just now.

eve-git:
the version update of sbc-common-components is in https://github.com/bcgov/sbc-common-components/pull/219. The two issues will be in the same release.

PR #221, opened 2021-05-31
Feature/Relationship #6921changes on Sbc-Header

karthik-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/6921

*Description of changes:*

- Updated Sbc-header component to accept a dashboardReturnUrl property that would be passed on while clicking "create account" to the Sbc-Auth so that can be used while displaying the Duplicate account warning page.
- Updated package version to 2.4.17

### DIFF
--- a/vue/sbc-common-components/package.json
+++ b/vue/sbc-common-components/package.json
@@ -1,6 +1,6 @@
 {
   "name": "sbc-common-components",
-  "version": "2.4.16",
+  "version": "2.4.17",
   "private": false,
   "description": "Common Vue Components to be used across BC Registries and Online Services.",
   "license": "Apache-2.0",
--- a/vue/sbc-common-components/src/components/SbcHeader.vue
+++ b/vue/sbc-common-components/src/components/SbcHeader.vue
@@ -455,14 +455,8 @@ export default class SbcHeader extends Mixins(NavigationMixin) {
   }
 
   private goToCreateBCSCAccount () {
-    if (this.inAuth) {
-      this.$router.push(`/${Pages.CREATE_ACCOUNT}`)
-    } else {
-      // To send the navigateBackUrl to duplicate account creator interceptor screen, so that we can redirect to that page.
-      let redirectUrl = new URL(`${ConfigHelper.getAuthContextPath()}/${Pages.CREATE_ACCOUNT}`)
-      if (this.dashboardReturnUrl) redirectUrl.searchParams.append('redirectToUrl', encodeURIComponent(this.dashboardReturnUrl))
-      window.location.assign(redirectUrl.toString())
-    }
+    const redirectUrl: string = this.dashboardReturnUrl ? `${Pages.CREATE_ACCOUNT}?redirectToUrl=${encodeURIComponent(this.dashboardReturnUrl)}` : Pages.CREATE_ACCOUNT
+    this.redirectToPath(this.inAuth, redirectUrl)
   }
 
   private async goToAccountInfo (settings: UserSettings) {
--- a/vue/sbc-common-components/src/components/SbcHeader.vue
+++ b/vue/sbc-common-components/src/components/SbcHeader.vue
@@ -374,6 +374,7 @@ export default class SbcHeader extends Mixins(NavigationMixin) {
   @Prop({ default: false }) inAuth!: boolean;
   @Prop({ default: false }) showProductSelector!: boolean;
   @Prop({ default: true }) showActions!: boolean;
+  @Prop({ default: '' }) dashboardReturnUrl !: string;
 
   private readonly loginOptions = [
     {
@@ -454,7 +455,14 @@ export default class SbcHeader extends Mixins(NavigationMixin) {
   }
 
   private goToCreateBCSCAccount () {
-    this.redirectToPath(this.inAuth, Pages.CREATE_ACCOUNT)
+    if (this.inAuth) {
+      this.$router.push(`/${Pages.CREATE_ACCOUNT}`)
+    } else {
+      // To send the navigateBackUrl to duplicate account creator interceptor screen, so that we can redirect to that page.
+      let redirectUrl = new URL(`${ConfigHelper.getAuthContextPath()}/${Pages.CREATE_ACCOUNT}`)
+      if (this.dashboardReturnUrl) redirectUrl.searchParams.append('redirectToUrl', encodeURIComponent(this.dashboardReturnUrl))
+      window.location.assign(redirectUrl.toString())
+    }
   }
 
   private async goToAccountInfo (settings: UserSettings) {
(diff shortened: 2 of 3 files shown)


vue/sbc-common-components/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "sbc-common-components",
-  "version": "2.4.16",
+  "version": "2.4.17",
```
severinbeauvais, line 3: Please also commit the package-lock.json file (so that the build server uses exactly the package versions that you used -- otherwise the build server will fetch the latest package versions, which may cause issues).

vue/sbc-common-components/src/components/SbcHeader.vue
```diff
@@ -454,7 +455,14 @@ export default class SbcHeader extends Mixins(NavigationMixin) {
...
   private goToCreateBCSCAccount () {
-    this.redirectToPath(this.inAuth, Pages.CREATE_ACCOUNT)
+    if (this.inAuth) {
+      this.$router.push(`/${Pages.CREATE_ACCOUNT}`)
+    } else {
+      // To send the navigateBackUrl to duplicate account creator interceptor screen, so that we can redirect to that page.
+      let redirectUrl = new URL(`${ConfigHelper.getAuthContextPath()}/${Pages.CREATE_ACCOUNT}`)
+      if (this.dashboardReturnUrl) redirectUrl.searchParams.append('redirectToUrl', encodeURIComponent(this.dashboardReturnUrl))
+      window.location.assign(redirectUrl.toString())
+    }
```
severinbeauvais, line 465: Can you please explain what impacts this has to the UIs that use this library. What flow(s) does this impact, and how? Do the UIs need to update anything, or provide new props?

severinbeauvais (commented):
I don't understand how this PR resolves ticket 6921. For example, that ticket describes 3 scenarios, each with a "looks like you already have an account" mockup, but this PR does not implement those mockups.

What design decision was made for this ticket? Please document it in the ticket and/or this PR.

vue/sbc-common-components/src/components/SbcHeader.vue
```diff
@@ -454,7 +455,14 @@ export default class SbcHeader extends Mixins(NavigationMixin) {
   }
 
   private goToCreateBCSCAccount () {
-    this.redirectToPath(this.inAuth, Pages.CREATE_ACCOUNT)
+    if (this.inAuth) {
```
shabeeb-aot, line 458: use the existing function this.redirectToPath if we can re-use

vue/sbc-common-components/package.json
karthik-aot: Thanks. I have pushed the package-lock.json file as well

vue/sbc-common-components/src/components/SbcHeader.vue
karthik-aot: Thanks, I have refactored the changes

vue/sbc-common-components/src/components/SbcHeader.vue
karthik-aot: Hi. As part of 6921. the idea is to intercept user with a duplicate account warning screen when the user already has an existing account and try to create another account. 
Therefore, when the user clicks create account, 

- In the sbc-common:(sbc-header) component, we capture the redirect url link ( as a property ) and direct to setup-account of sbc-auth. 
- In the sbc-auth, when the user decides not to create account and tries to access existing account, we route to the redirect url that we captured earlier.
- A default url is also used incase there is no redirect url. ( sbc-auth home)
- For you reference, we have created a dependency ticket as well https://github.com/bcgov/entity/issues/7757
- And a PR in sbc-auth as well for your reference https://github.com/bcgov/sbc-auth/pull/1646

severinbeauvais: So the other UIs (Filings UIs, Create UI, Edit UI, Namerequest UI and PPR UI) don't need to change anything (except to import the latest sbc-common-components) ?

karthik-aot: And to pass the redirect url in the **dashboardReturnUrl** property in the Sbc-Header value so that we can direct the user to that link when the user selects access account button.
image

severinbeauvais: Are you saying each UI needs to pass in this new prop?

Could this prop be defaulted from the current session storage variables (so that each UI does not have to populate the common variables)?

karthik-aot: we can pass the dashboardReturnUrl as a property as follows

```
<sbc-header
        :key="$store.state.refreshKey"
        :in-auth="true"
        :show-product-selector="false"
        :dashboardReturnUrl="dashboardReturnUrl"
        @account-switch-started="startAccountSwitch"
        @account-switch-completed="completeAccountSwitch"
        @hook:mounted="setup"
        ref="header" :redirect-on-logout="logoutUrl">
        <template v-slot:login-button-text>
          Log in with BC Services Card
        </template>
      </sbc-header>
```

saravanpa-aot (approved):

vue/sbc-common-components/src/components/SbcHeader.vue
severinbeauvais: OK, thanks.

-> Do the UIs need to pass this new prop?

karthik-aot: We are defaulting it to Sbc-auth /home route incase there is no prop passed, but we can pass the value as well :)

severinbeauvais: I still don't understand if the UIs (Filings, Namerequest, etc) are **required** to pass in the new prop or not.

saravanpa-aot: Hi @severinbeauvais the dependency ticket https://github.com/bcgov/entity/issues/7757 will come to your way soon... This is not a mandatory param , If you need any behaviour other than the default behaviour , the url could be passed in as a prop. That decision will be a discussion with in each team .
So its not required , But if you want customization , you can pass the property.
This is not an every day behaviour which user will going through , rather this is a preventive measure incase if user accidentally or unintentionally clicks on create account OR somehow has a hard time finding the current account ..

severinbeauvais: Thanks Saravan. That's exactly what I wanted to know 👍

severinbeauvais:
> I don't understand how this PR resolves ticket 6921. For example, that ticket describes 3 scenarios, each with a "looks like you already have an account" mockup, but this PR does not implement those mockups.
> 
> What design decision was made for this ticket? Please document it in the ticket and/or this PR.

Saaravan says this is OK.

severinbeauvais (approved):

shabeeb-aot (approved):

PR #228, opened 2021-06-10
1password keys integration.

pwei1018:
*Issue #:*
https://github.com/bcgov/entity/issues/7849
https://github.com/bcgov/entity/issues/6801

*Description of changes:*
1. change the key new to match new configuration of 1password.
2. change to read the session keys directly instead of read from auth-api-config session key.

### DIFF
--- a/vue/sbc-common-components/package.json
+++ b/vue/sbc-common-components/package.json
@@ -1,6 +1,6 @@
 {
   "name": "sbc-common-components",
-  "version": "2.4.18",
+  "version": "2.4.19",
   "private": false,
   "description": "Common Vue Components to be used across BC Registries and Online Services.",
   "license": "Apache-2.0",
--- a/vue/sbc-common-components/src/util/constants.ts
+++ b/vue/sbc-common-components/src/util/constants.ts
@@ -7,6 +7,8 @@ export enum SessionStorageKeys {
   LaunchDarklyFlags = 'LD_FLAGS',
   CurrentAccount = 'CURRENT_ACCOUNT',
   AuthApiUrl = 'AUTH_API_URL',
+  AuthWebUrl = 'AUTH_WEB_URL',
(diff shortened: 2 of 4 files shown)


vue/sbc-common-components/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "sbc-common-components",
-  "version": "2.4.18",
+  "version": "2.4.19",
```
severinbeauvais, line 3: The package-lock file needs to be checked in also.

vue/sbc-common-components/src/util/constants.ts
```diff
@@ -7,6 +7,8 @@ export enum SessionStorageKeys {
   LaunchDarklyFlags = 'LD_FLAGS',
   CurrentAccount = 'CURRENT_ACCOUNT',
   AuthApiUrl = 'AUTH_API_URL',
+  AuthWebUrl = 'AUTH_WEB_URL',
+  StatusApiUrl = 'STATUS_API_URL',
```
severinbeauvais, line 11: Where is this used?

vue/sbc-common-components/package.json
pwei1018: updated

vue/sbc-common-components/src/util/constants.ts
pwei1018: use in config-helper. I make the code looks like consistence.

severinbeauvais: I see it now 👍

severinbeauvais (changes requested):
~Something went wrong with package-lock.json update.~ _now fixed_

severinbeauvais (approved):

karthik-aot (approved):

severinbeauvais (approved):

shabeeb-aot (approved):

PR #230, opened 2021-06-21
Added skippable urls from pending orgs

saravanpa-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/7781

*Description of changes:*

- some urls are allowed for pending orgs so that user can submit the account details again
- list of urls are given which is checked against the target url

The approach of keeping this url can get complicated since it creates multiple redirections on refreshes. So not resorting to that.

### DIFF
--- a/vue/sbc-common-components/src/components/SbcHeader.vue
+++ b/vue/sbc-common-components/src/components/SbcHeader.vue
@@ -489,7 +489,7 @@ export default class SbcHeader extends Mixins(NavigationMixin) {
       const targetPath = window.location.pathname
       const substringCheck = (element:string) => targetPath.indexOf(element) > -1
       // check if any of the url is the allowed uri
-      let isAllowedUrl = ALLOWED_URIS_FOR_PENDING_ORGS.findIndex(substringCheck) > -1
+      const isAllowedUrl = ALLOWED_URIS_FOR_PENDING_ORGS.findIndex(substringCheck) > -1
       if (!isAllowedUrl) {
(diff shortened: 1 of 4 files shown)


sumesh-aot (approved):

vue/sbc-common-components/src/components/SbcHeader.vue
```diff
@@ -486,8 +486,14 @@ export default class SbcHeader extends Mixins(NavigationMixin) {
     if ([AccountStatus.NSF_SUSPENDED, AccountStatus.SUSPENDED].some(status => status === this.currentAccount?.accountStatus)) {
       this.redirectToPath(this.inAuth, `${Pages.ACCOUNT_FREEZ}`)
     } else if (this.currentAccount?.accountStatus === AccountStatus.PENDING_STAFF_REVIEW) {
-      const accountName = encodeURIComponent(btoa(this.accountName))
-      this.redirectToPath(this.inAuth, `${Pages.PENDING_APPROVAL}/${accountName}/true`)
+      const targetPath = window.location.pathname
+      const substringCheck = (element:string) => targetPath.indexOf(element) > -1
+      // check if any of the url is the allowed uri
+      let isAllowedUrl = ALLOWED_URIS_FOR_PENDING_ORGS.findIndex(substringCheck) > -1
```
shabeeb-aot, line 492: can use cont instead of let ?

shabeeb-aot (approved):

shabeeb-aot:
update version in package-lock to reflect both PR

PR #246, opened 2021-12-16
9811 What's New notification panel

pwei1018:
*Issue #:*
https://github.com/bcgov/entity/issues/9811

*Description of changes:*
Added What's New notification panel.

### DIFF
--- a/vue/sbc-common-components/package.json
+++ b/vue/sbc-common-components/package.json
@@ -1,6 +1,6 @@
 {
   "name": "sbc-common-components",
-  "version": "2.4.28",
+  "version": "2.4.29",
   "private": false,
   "description": "Common Vue Components to be used across BC Registries and Online Services.",
   "license": "Apache-2.0",
@@ -21,6 +21,7 @@
   "dependencies": {
     "@mdi/font": "^4.5.95",
     "axios": "^0.21.1",
+    "clickout-event": "^1.1.2",
     "core-js": "^3.1.4",
     "country-list": "^2.2.0",
     "keycloak-js": "^9.0.3",
--- a/vue/sbc-common-components/src/components/NotificationPanel.vue
+++ b/vue/sbc-common-components/src/components/NotificationPanel.vue
@@ -8,12 +8,10 @@
       right
       app
       :width="440"
-      style="top: 70px; max-height: calc(100% - 0);"
     >
       <v-app-bar
         flat
         outlined
-        color="#FCBA19"
       >
         <v-toolbar-title class="toolbar-title">What's New at BC Registries</v-toolbar-title>
         <v-spacer></v-spacer>
@@ -103,8 +101,13 @@ export default class NotificationPanel extends Vue {
 </script>
 
 <style lang="scss" scoped>
+@import "../assets/scss/layout.scss";
 @import "../assets/scss/theme.scss";
 
+//$app-notification-height: calc(100vh - $app-header-height);
+//$app-notification-item-height: $app-notification-height/3;
+
+//@debug $app-notification-item-height;
 ::v-deep ::-webkit-scrollbar {
   width: 2px;
 }
@@ -114,9 +117,13 @@ export default class NotificationPanel extends Vue {
   border-radius: 20px;
 }
 
-.navigation-drawer {
-  top: 100px;
-  max-height: calc(100% - 0%);
+::v-deep .v-navigation-drawer--right {
+  transform: translatey($app-header-height) !important;
+  height: 100vh;
+}
+
+.v-app-bar.v-toolbar.v-sheet {
+  background-color: $app-notification-orange !important;
 }
 
 .dialog-close {
@@ -129,7 +136,7 @@ export default class NotificationPanel extends Vue {
 }
 
 ::v-deep .v-btn:not(.dialog-close) .v-icon.v-icon {
-  font-size: 1.25rem !important;
+  font-size: $px-18 !important;
 }
 
 ::v-deep .v-btn__content {
@@ -138,25 +145,19 @@ export default class NotificationPanel extends Vue {
 
 .toolbar-title {
   color: $BCgovBlue5;
-  font-size: 1.125rem;
+  font-size: $px-18;
   font-weight: 700;
 }
 
 .list-subtitle {
-  font-size: 1.125rem;
+  font-size: $px-18;
   font-weight: 700;
-}
-
-.list-text {
-  line-height: 1;
-  word-spacing: .05em;
-  letter-spacing: .01em;
 }
 
 .dot-red {
   height: 8px;
   width: 8px;
-  background-color: #F44336;
+  background-color: $app-notification-red !important;
   border-radius: 50%;
   display: inline-block;
   margin-top: 18px;
@@ -164,12 +165,7 @@ export default class NotificationPanel extends Vue {
 }
 
 .dot-blue {
-  height: 8px;
-  width: 8px;
-  background-color: #2196F3;
-  border-radius: 50%;
-  display: inline-block;
-  margin-top: 18px;
-  margin-left: 10px;
+  @extend .dot-red;
+  background-color: $app-notification-blue !important;
 }
 </style>
--- a/vue/sbc-common-components/src/components/NotificationPanel.vue
+++ b/vue/sbc-common-components/src/components/NotificationPanel.vue
@@ -1,0 +1,175 @@
+<template>
+  <div
+    v-if="showNotifications"
+    v-on:clickout="emitClose()"
+  >
+    <v-overlay></v-overlay>
+    <v-navigation-drawer
+      right
+      app
+      :width="440"
+      style="top: 70px; max-height: calc(100% - 0);"
+    >
+      <v-app-bar
+        flat
+        outlined
+        color="#FCBA19"
+      >
+        <v-toolbar-title class="toolbar-title">What's New at BC Registries</v-toolbar-title>
+        <v-spacer></v-spacer>
+        <v-btn
+          icon
+          large
+          class="dialog-close"
+          @click="emitClose()"
+        >
+          <v-icon>mdi-close</v-icon>
+        </v-btn>
+      </v-app-bar>
+      <v-list flat>
+        <v-list-item-group color="primary">
+          <template v-for="(item, i) in notifications">
+            <v-list-item :key="i">
+              <v-row dense>
+                <v-col class="d-flex" cols="1">
+                  <span :class="!item.read && (item.priority ? 'dot-red' : 'dot-blue')">
+                  </span>
+                </v-col>
+                <v-col>
+                <v-list-item-content>
+                  <v-list-item-title class="font-weight-bold list-subtitle">{{item.title}}</v-list-item-title>
+                  <v-list-item-subtitle>{{item.date}}</v-list-item-subtitle>
+                  <v-spacer></v-spacer>
+                  <v-list-item-content v-html="item.description"></v-list-item-content>
+                </v-list-item-content>
+                </v-col>
+              </v-row>
+            </v-list-item>
+            <v-divider v-if="i < notifications.length - 1" :key="`${i}-divider`"></v-divider>
+          </template>
+        </v-list-item-group>
+      </v-list>
+    </v-navigation-drawer>
+
+  </div>
+</template>
+
+<script lang='ts'>
+import { Component, Emit, Prop, Vue } from 'vue-property-decorator'
+import { Notification } from '../models/notification'
+import { mapState, mapActions } from 'vuex'
+import NotificationModule from '../store/modules/notification'
+import { getModule } from 'vuex-module-decorators'
+import 'clickout-event'
+
+@Component({
+  name: 'NotificationPanel',
+  beforeCreate () {
+    this.$store.isModuleRegistered = function (aPath: string[]) {
+      let m = (this as any)._modules.root
+      return aPath.every((p) => {
+        m = m._children[p]
+        return m
+      })
+    }
+    if (!this.$store.isModuleRegistered(['notification'])) {
+      this.$store.registerModule('notification', NotificationModule)
+    }
+    this.$options.computed = {
+      ...(this.$options.computed || {}),
+      ...mapState('notification', ['notifications'])
+    }
+    this.$options.methods = {
+      ...(this.$options.methods || {}),
+      ...mapActions('notification', ['markAsRead'])
+    }
+  }
+})
+export default class NotificationPanel extends Vue {
+  private readonly notifications!: Notification[]
+
+  /** Prop to display the dialog. */
+  @Prop() showNotifications: boolean
+
+  @Emit('closeNotifications')
+  private async emitClose () {
+
+  }
+
+  private async mounted () {
+    getModule(NotificationModule, this.$store)
+  }
+}
+</script>
+
+<style lang="scss" scoped>
+@import "../assets/scss/theme.scss";
+
+::v-deep ::-webkit-scrollbar {
+  width: 2px;
+}
+
+::v-deep ::-webkit-scrollbar-thumb {
+  background: black;
+  border-radius: 20px;
+}
+
+.navigation-drawer {
+  top: 100px;
+  max-height: calc(100% - 0%);
+}
+
+.dialog-close {
+  position: absolute;
+  top: 8px;
+  right: 15px;
+  margin-right: 70px;
+  z-index: 2;
+  font-weight: bold;
+}
+
+::v-deep .v-btn:not(.dialog-close) .v-icon.v-icon {
+  font-size: 1.25rem !important;
+}
+
+::v-deep .v-btn__content {
+  line-height: inherit;
+}
+
+.toolbar-title {
+  color: $BCgovBlue5;
+  font-size: 1.125rem;
+  font-weight: 700;
+}
+
+.list-subtitle {
+  font-size: 1.125rem;
+  font-weight: 700;
+}
+
+.list-text {
+  line-height: 1;
+  word-spacing: .05em;
+  letter-spacing: .01em;
+}
+
+.dot-red {
+  height: 8px;
+  width: 8px;
+  background-color: #F44336;
+  border-radius: 50%;
+  display: inline-block;
+  margin-top: 18px;
+  margin-left: 10px;
+}
+
+.dot-blue {
+  height: 8px;
+  width: 8px;
+  background-color: #2196F3;
+  border-radius: 50%;
+  display: inline-block;
+  margin-top: 18px;
+  margin-left: 10px;
+}
+</style>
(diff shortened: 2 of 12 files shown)


vue/sbc-common-components/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "sbc-common-components",
-  "version": "2.4.28",
+  "version": "2.4.29",
```
severinbeauvais, line 3: Please run `npm i` so that the package lock file reflects this version change.

vue/sbc-common-components/src/components/NotificationPanel.vue
```diff
@@ -0,0 +1,175 @@
...
+  <div
+    v-if="showNotifications"
+    v-on:clickout="emitClose()"
+  >
+    <v-overlay></v-overlay>
+    <v-navigation-drawer
+      right
+      app
+      :width="440"
+      style="top: 70px; max-height: calc(100% - 0);"
```
severinbeauvais, line 11: How does this look when banners are displayed?

Could `top` be calculated based on the sbc-header height (in CSS maybe)?

pwei1018: How to calculate it?

vue/sbc-common-components/src/components/NotificationPanel.vue
```diff
@@ -0,0 +1,175 @@
...
+    >
+      <v-app-bar
+        flat
+        outlined
+        color="#FCBA19"
```
severinbeauvais, line 16: It's better to not have hex colour values in components.

You can use a theme colour name here, since one was declared here:
https://github.com/bcgov/sbc-common-components/blob/master/vue/sbc-common-components/src/plugins/vuetify.ts

vue/sbc-common-components/src/components/NotificationPanel.vue
```diff
@@ -0,0 +1,175 @@
...
+                <v-list-item-content>
+                  <v-list-item-title class="font-weight-bold list-subtitle">{{item.title}}</v-list-item-title>
+                  <v-list-item-subtitle>{{item.date}}</v-list-item-subtitle>
+                  <v-spacer></v-spacer>
+                  <v-list-item-content v-html="item.description"></v-list-item-content>
+                </v-list-item-content>
+                </v-col>
+              </v-row>
+            </v-list-item>
+            <v-divider v-if="i < notifications.length - 1" :key="`${i}-divider`"></v-divider>
```
severinbeauvais, line 48: Another time, you can simply render this divider here, and then use CSS (last-of-type or similar) to hide the last divider 😀

vue/sbc-common-components/src/components/NotificationPanel.vue
```diff
@@ -0,0 +1,175 @@
...
+import { Component, Emit, Prop, Vue } from 'vue-property-decorator'
+import { Notification } from '../models/notification'
+import { mapState, mapActions } from 'vuex'
+import NotificationModule from '../store/modules/notification'
+import { getModule } from 'vuex-module-decorators'
+import 'clickout-event'
+
+@Component({
+  name: 'NotificationPanel',
+  beforeCreate () {
```
severinbeauvais, line 67: I recommend moving this method inside the class body, so that the Component object remains as clean as possible.

vue/sbc-common-components/src/components/NotificationPanel.vue
severinbeauvais, line 11: It looks like you have CSS for these in the style block below, so can you delete this property?

vue/sbc-common-components/src/components/NotificationPanel.vue
```diff
@@ -0,0 +1,175 @@
...
+
+.list-subtitle {
+  font-size: 1.125rem;
+  font-weight: 700;
+}
+
+.list-text {
+  line-height: 1;
+  word-spacing: .05em;
+  letter-spacing: .01em;
```
severinbeauvais, line 153: We usually use rems instead of ems. I think they're mostly compatible here.

vue/sbc-common-components/src/components/NotificationPanel.vue
pwei1018: The CSS one doesn't work. So I put it to give a try. I will look at it.

vue/sbc-common-components/src/components/NotificationPanel.vue
```diff
@@ -0,0 +1,175 @@
...
+}
+
+.dot-red {
+  height: 8px;
+  width: 8px;
+  background-color: #F44336;
```
severinbeauvais, line 159: Please use a variable for this colour. If needed, you can define a new variable in theme.scss.

Ditto for lline 169 below.

vue/sbc-common-components/src/components/NotificationPanel.vue
```diff
@@ -0,0 +1,175 @@
...
+  font-size: 1.25rem !important;
+}
+
+::v-deep .v-btn__content {
+  line-height: inherit;
+}
+
+.toolbar-title {
+  color: $BCgovBlue5;
+  font-size: 1.125rem;
```
severinbeauvais, line 141: In Entity UIs, we've started using variables for font sizes (to keep things manageable). I recommend it.

Eg, https://github.com/bcgov/business-filings-ui/blob/4f515bf90998caa80b47fe8bf8abcfba94209e36/src/assets/styles/theme.scss#L71

vue/sbc-common-components/src/components/NotificationPanel.vue
pwei1018: example?

vue/sbc-common-components/src/components/NotificationPanel.vue
```diff
@@ -0,0 +1,175 @@
...
+  font-weight: 700;
+}
+
+.list-subtitle {
+  font-size: 1.125rem;
+  font-weight: 700;
```
severinbeauvais, line 147: Since we don't have several levels of bolding, some of us have been using just 'normal' and 'bold' for font weights. In fact, you use 'bold' on line 128 above.

vue/sbc-common-components/src/components/NotificationPanel.vue
```diff
@@ -0,0 +1,175 @@
...
+  width: 8px;
+  background-color: #F44336;
+  border-radius: 50%;
+  display: inline-block;
+  margin-top: 18px;
+  margin-left: 10px;
+}
+
+.dot-blue {
+  height: 8px;
```
severinbeauvais, line 167: Another time, you could use `@extend` to inherit all properties from .dot-red and just override the colour.

Example: https://github.com/bcgov/business-filings-ui/blob/4f515bf90998caa80b47fe8bf8abcfba94209e36/src/assets/styles/base.scss#L76

vue/sbc-common-components/src/components/SbcHeader.vue
```diff
@@ -20,6 +20,31 @@
...
+              overlap
+              offset-y="-5"
+              offset-x="10"
+              color="notificationUnreadPriorityCount > 0 ? error : blue"
+              v-if="notificationUnreadCount > 0"
+            >
+            </v-badge>
+            <span>
+              What's New
+            </span>
```
severinbeauvais, line 45: The span does nothing for you here.

vue/sbc-common-components/src/components/SbcHeader.vue
```diff
@@ -20,6 +20,31 @@
...
+            attach="#appHeader"
+            @click.stop="notificationPanel=true"
+            v-if="!isAuthenticated && notificationCount > 0 && isWhatsNewOpen"
+          >
+            <v-badge
+              dot
+              overlap
+              offset-y="-5"
+              offset-x="10"
+              color="notificationUnreadPriorityCount > 0 ? error : blue"
```
severinbeauvais, line 39: Does this work?
1. I'm surprised you don't need to use `:color=`
2. I'm surprised you don't need to use `'error' : 'blue'`

vue/sbc-common-components/src/components/SbcHeader.vue
```diff
@@ -437,6 +488,11 @@ export default class SbcHeader extends Mixins(NavigationMixin) {
       // checking for account status
       await this.checkAccountStatus()
     }
+
+    await this.syncNotifications()
+    await this.fetchNotificationCount()
+    await this.fetchNotificationUnreadPriorityCount()
+    await this.fetchNotificationUnreadCount()
```
severinbeauvais, line 495: I'm wondering if these could be done in the background so they don't slow down the loading of the sbc header...?

vue/sbc-common-components/src/models/notification.ts
```diff
@@ -0,0 +1,26 @@
...
+  expiryDate?: string;
+  labels?: NotificationLabel[];
+  title: string;
+  description: string;
+  app: AppType;
+  priority: boolean;
+  read: boolean;
+}
+
+export type Notifications = Notification[];
```
severinbeauvais, line 26: Sometimes lint requires a blank line at the end of a file.

vue/sbc-common-components/src/services/notification.services.ts
```diff
@@ -0,0 +1,14 @@
...
+import Axios, { AxiosResponse } from 'axios'
+import ConfigHelper from '../util/config-helper'
+import { addAxiosInterceptors } from '../util/interceptors'
+import { Notifications } from '../models/notification'
+
+const axios = addAxiosInterceptors(Axios.create())
+
+export default class NotificationService {
+  static async getNotifications (): Promise<AxiosResponse<Notifications>> {
+    console.log('getNotifications')
```
severinbeauvais, line 11: Remove debugging code some time.

vue/sbc-common-components/src/store/modules/notification.ts
```diff
@@ -0,0 +1,77 @@
...
+
+@Module({
+  name: 'notification',
+  namespaced: true
+})
+export default class NotificationModule extends VuexModule {
+  notifications: Notifications = []
+  notificationCount = 0
+  notificationUnreadPriorityCount = 0
+  notificationUnreadCount = 0
```
severinbeauvais, line 15: It would be better to use getters instead of accessing a module's state directly (as you do in SbcHeader.vue).

But, this project uses state a lot, so 🤷‍♂️

vue/sbc-common-components/src/store/modules/notification.ts
```diff
@@ -0,0 +1,77 @@
...
+  @Action({ rawError: true, commit: 'setNotificationUnreadCount' })
+  public async fetchNotificationUnreadCount (): Promise<number> {
+    return this.notifications.filter(notification => !notification.read).length
+  }
+
+  @Action({ rawError: true, commit: 'setNotificationUnreadCount' })
+  public async markAsRead (): Promise<Notifications> {
+    let nl = JSON.parse(ConfigHelper.getFromSession(SessionStorageKeys.WhatsNew) || '{}')
+    nl.map(notification => { notification.read = true; return notification })
+    console.log('markAsRead: ' + nl)
```
severinbeauvais, line 73: Remove debugging code some time.

vue/sbc-common-components/src/store/modules/notification.ts
```diff
@@ -0,0 +1,77 @@
...
+  @Mutation
+  public setNotificationUnreadCount (count: number): void {
+    this.notificationUnreadCount = count
+  }
+
+  @Action({ rawError: true, commit: 'setNotifications' })
+  public async syncNotifications (): Promise<Notifications> {
+    const response = await NotificationService.getNotifications()
+    if (response && response.data) {
+      console.log('notification', response.data)
```
severinbeauvais, line 42: Remove debugging code some time.

vue/sbc-common-components/tests/unit/notification.services.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+    // @ts-ignore
+    Axios.spread.mockReturnValue(mockAxiosSpreadResult)
+    sessionStorage.__STORE__[SessionStorageKeys.ApiConfigKey] = JSON.stringify({ VUE_APP_STATUS_ROOT_API: API_URL })
+    NotificationServices.getNotifications()
+  })
+
+  it('should call Axios.get once for status ', () => {
+    expect(Axios.get).toHaveBeenCalledWith(`${API_URL}/whatsnew`)
+  })
+})
```
severinbeauvais, line 33: How about some tests that the component displays various notifications as expected? Including read and not-read, etc.

vue/sbc-common-components/src/components/SbcHeader.vue
pwei1018: How?

vue/sbc-common-components/src/models/notification.ts
pwei1018: I added a blank line but my editor report linting error.

vue/sbc-common-components/src/components/NotificationPanel.vue
severinbeauvais: It might be possible to have a CSS `calc()` add up all the pieces that make up the header that you are trying to fit below.

Or, programatically, you could query various element heights, add them up, and dynamically set your panel's top.

This might be too complicated. But I really would like to see some sort of "$sbc-header-height" variable (constant) that is declared at an appropriate place (some main SCSS file) and that components such as this one can just use. This means only a single place will need to be updated when the header height changes.

severinbeauvais: If the v-navigation-drawer divs and classes are rendered at run-time, you may need to use `::v-deep` to select the right element.

vue/sbc-common-components/src/components/SbcHeader.vue
pwei1018: This is working. :) But I will change it as your recommended.

vue/sbc-common-components/src/components/NotificationPanel.vue
severinbeauvais: https://github.com/bcgov/namerequest/blob/9a6ae051254e97a29e75a38df947b140df3b0627/src/components/dialogs/receipts.vue#L99

severinbeauvais: ^^ That is if all the elements of the same type are together. You may need to use last-child or something else. Google it -- there are lots of examples.

vue/sbc-common-components/src/components/SbcHeader.vue
severinbeauvais: One way is to not await them -- assuming they are not dependents on each other -- and if you do this, add a comment stating that you are intentionally not awaiting an async method (and why you're doing it).

vue/sbc-common-components/src/models/notification.ts
severinbeauvais: Hahaha, OK, never mind!

Who knows the mysteries of Lint ?

vue/sbc-common-components/src/store/modules/notification.ts
severinbeauvais: You can probably ignore this conversation. I see that SbcHeader.vue uses both `mapState()` and `mapGetters()` to access variables in the different store modules, so I guess it doesn't matter one way or the other which you use. (The code pattern has already been set in this case.)

vue/sbc-common-components/src/plugins/vuetify.ts
```diff
@@ -54,8 +54,10 @@ export default new Vuetify({
         },
         bcgovgold: {
           base: '#fcba19'
-        }
+        },
+        appNotificationOrange: '#FCBA19'
```
severinbeauvais, line 58: Yes, this is the way to declare new colour variables that you can use in Vuetify.

But are you using this anywhere?

pwei1018: I didn't use this. I have it defined in theme.scss . Very confused about scss variables. Some are defined in vuetify.ts, most are defined in scss. Is there any standard for this?

severinbeauvais: In theme.scss it's an actual _CSS variable_ that you can use in other CSS.

In this file are the _Vuetify colour variables_ used by Vuetify components.

For example, instead of doing this:
`<v-btn color="#FCBA19">`

You can do this:
`<v-btn color="appNotificationOrange">`

In the example above, "color" is not a style/CSS, it's a Vuetify component prop.

As for a standard, I like to have no hex colours in the components -- named colours (or variables) only. Therefore for some elements we need to use CSS variables, and for some Vuetify components we need to use Vuetify colour variables. I try to name them the same, and also usually leave a comment on them, like this: [appNotificationOrange](https://github.com/bcgov/business-filings-ui/blob/42e669aff96183e900ceed1432dff0c2687edeee/src/main.ts#L74)

pwei1018: I tried to do ```<v-app-bar color="appNotificationOrange">```. But it doesn't work and so I put it in css 

```
.v-app-bar.v-toolbar.v-sheet {
  background-color: $app-notification-orange !important;
}
```

severinbeauvais: Sure.

vue/sbc-common-components/src/components/NotificationPanel.vue
pwei1018: I will ignore this recommendation and revisit it later when I have time.

vue/sbc-common-components/src/components/SbcHeader.vue
pwei1018: All these function tight together have to wait. But it should not cause performance issue. I will put the comment in it.

vue/sbc-common-components/tests/unit/notification.services.spec.ts
pwei1018: This is actually a draft version and will have more function to add it in. I will move the api call to notify service and add the test cases later.

vue/sbc-common-components/src/components/NotificationPanel.vue
severinbeauvais: As agreed, yes.

vue/sbc-common-components/src/components/SbcHeader.vue
severinbeauvais: OK, I'm glad you thought about it.

vue/sbc-common-components/tests/unit/notification.services.spec.ts
severinbeauvais: ok

severinbeauvais (approved):
Approved with expectation that test cases will come later.

PR #260, opened 2022-01-31
 Account may change when navigating between UIs#6151 | banner adding …

shabeeb-aot:
…account id

*Issue #:*
https://github.com/bcgov/entity/issues/6151

*Description of changes:*


karthik-aot (approved):

sumesh-aot (approved):

severinbeauvais:
When changes like this are made to the sbc-common-components, please ensure there are dependency tickets created for the affected apps (since they need to update their import version of this library).

PR #281, opened 2022-05-26
add extra filingLabel prop for sbcfeesummary component

meganawong:
*Issue #:*
https://github.com/bcgov/entity/issues/11660

*Description of changes:*
Added extra filing label for GP/SP in fee summary component.


seeker25 (approved):
Looks good to me.

vue/sbc-common-components/src/components/SbcFeeSummary.vue
```diff
@@ -16,7 +16,7 @@
         <li class="container fee-list__item"
           :key="lineItem.filingType"
           >
-          <div class="fee-list__item-name">{{lineItem.filingType}}</div>
+          <div class="fee-list__item-name">{{filingLabel ? filingLabel : lineItem.filingType}}</div>
```
severinbeauvais, line 19: FYI, the other way to do this is `filingLabel || lineItem.filingType`.

But this is fine.

vue/sbc-common-components/src/components/SbcFeeSummary.vue
```diff
@@ -71,6 +71,9 @@ export default class SbcFeeSummary extends Vue {
   @Prop({ default: '' })
   private payURL!: string
 
+  @Prop()
+  private filingLabel!: string
```
severinbeauvais, line 75: FYI, it's usually a good idea to provide a default value (as on line 71) so it's never undefined.

If you're concerned about being able to identify an empty string, all of these are "falsy":
- filingLabel is undefined
- filingLabel is null
- filingLabel is ''

That's why you can use use a simple expression to check if filingLabel is valid.

No change needed.

vue/sbc-common-components/src/components/SbcFeeSummary.vue
severinbeauvais: This implementation sort of falls down is there are multiple line items (they will all use the override label).

Probably OK for now but may need to be fixed in the future if there are multiple fees and we need to override a fee label.

PR #289, opened 2022-06-16
 AUTH-WEB / COMMON-COMPONENTS - User Navigation while switching accou…

shabeeb-aot:
…nts#c

*Issue #:*
https://github.com/bcgov/entity/issues/12363

*Description of changes:*
* User redirection on switch account

### DIFF
--- a/vue/sbc-common-components/src/components/SbcHeader.vue
+++ b/vue/sbc-common-components/src/components/SbcHeader.vue
@@ -601,6 +601,11 @@ export default class SbcHeader extends Mixins(NavigationMixin) {
     const currentURL = trimTrailingSlashURL(`${window.location.origin}${window.location.pathname}`)
     const redirect = allowedRedirectURls.indexOf(currentURL) > -1 ? currentURL : defaultRedirect
 
+    // @Prop({ default: false }) skipAccountSwitchRedirect!: boolean;
+
+    // skipAccountSwitchRedirect as prop and check here if need to avoid redirect to dashboard
+    // handle all the condtion (like NFS/pending approval page) in own app when we are doing it
+
     if (!inAuth) {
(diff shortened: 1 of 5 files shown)


vue/sbc-common-components/src/components/SbcHeader.vue
```diff
@@ -590,8 +592,17 @@ export default class SbcHeader extends Mixins(NavigationMixin) {
...
+    // check for allowed redirect to determine whether need to redirect back to that page or dashboard
+    // list of Allowed URLs
+    const allowedRedirectURls = ConfigHelper.getAllowedUrlForRedirectToSamePage()
+    // default redirect to dashboard
+    const defaultRedirect = `${ConfigHelper.getRegistryHomeURL()}/${Pages.REGISTRY_DASHBOARD}`
+
+    const currentURL = trimTrailingSlashURL(`${window.location.origin}${window.location.pathname}`)
+    const redirect = allowedRedirectURls.indexOf(currentURL) > -1 ? currentURL : defaultRedirect
+
     if (!inAuth) {
```
seeker25, line 604: I think for the future it would be nice if we allowed another prop condition here. 

Maybe not `inAuth`, but something like `skipAccountSwitchRedirect` (default false)? 

That way in the future various applications could handle the 
`'account-switch-completed'` event if they needed and bypass the re-direct?

seeker25:
Might want to update the package.json version as well.

seeker25 (approved):

vue/sbc-common-components/src/components/SbcHeader.vue
shabeeb-aot: commented in code now, In future when someone wants to use, they have made changes in applications also, Then we can add this prop. Now if we gave it as a prop i think it will skip all the checks like NFS, Account pending

PR #297, opened 2022-07-21
Update to version 3.0.0 to be consistent with other projects.

seeker25:
Upgrade package version to be consistent with other projects.


shabeeb-aot (approved):

severinbeauvais:
FYI, there is no consistency with other projects. Their versions were all updated independently depending on the features that went in them.

We will probably go to the next version when we do the Vue/Vuetify upgrades soon. cc: @pwei1018

PR #304, opened 2022-10-05
13825 - Add in ? for replacing a url, this way it can return undefine…

seeker25:
…d or empty instead of throwing an error.

*Issue #:*
https://github.com/bcgov/entity/issues/13825

### DIFF
--- a/vue/sbc-common-components/src/util/common-util.ts
+++ b/vue/sbc-common-components/src/util/common-util.ts
@@ -66,6 +66,6 @@ export function appendAccountId (url, accountId = '') {
 
 export function removeAccountIdFromUrl (url, key = ACCOUNT_ID) {
   // replacing account id and formattig ie, removing extra ? or &
-  return url.replace(new RegExp(key + '=\\w+'), '').replace('?&', '?').replace(/\?$/, '')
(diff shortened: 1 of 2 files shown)


vue/sbc-common-components/src/util/common-util.ts
```diff
@@ -66,6 +66,6 @@ export function appendAccountId (url, accountId = '') {
 
 export function removeAccountIdFromUrl (url, key = ACCOUNT_ID) {
   // replacing account id and formattig ie, removing extra ? or &
-  return url.replace(new RegExp(key + '=\\w+'), '').replace('?&', '?').replace(/\?$/, '')
-    .replace('&&', '&').replace(/&$/, '')
+  return url?.replace(new RegExp(key + '=\\w+'), '')?.replace('?&', '?')?.replace(/\?$/, '')
+    ?.replace('&&', '&')?.replace(/&$/, '')
```
severinbeauvais, line 70: Mind. Blown.

lol

severinbeauvais (approved):

vue/sbc-common-components/src/util/common-util.ts
seeker25: I think that's right.. probably only need `url?.`

PR #311, opened 2022-10-17
13914 Added CSS to BaseAddress

severinbeauvais:
*Issue #:* bcgov/entity#13914

*Description of changes:*
- added CSS to BaseAddress (so that apps don't have to override these basic styles)
- app version = 3.0.4

### DIFF
--- a/vue/sbc-common-components/src/components/BaseAddress.vue
+++ b/vue/sbc-common-components/src/components/BaseAddress.vue
@@ -472,6 +472,10 @@ export default class BaseAddress extends Mixins(ValidationMixin, CountriesProvin
   flex: 1 1 auto;
 }
 
+.address-block__info-row {
+  color: $gray7;
+}
+
 // Form Row Elements
 .form__row.three-column {
   align-items: stretch;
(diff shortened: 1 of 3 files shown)


vue/sbc-common-components/src/components/BaseAddress.vue
```diff
@@ -488,6 +492,20 @@ export default class BaseAddress extends Mixins(ValidationMixin, CountriesProvin
...
+  font-weight: normal;
+}
+
+// text field inputs
+::v-deep {
+  .v-input input {
+    color: $gray9;
+  }
+}
+
```
severinbeauvais, line 508: If these base styles aren't set then the elements inherit whatever Vuetify provides, which is not what our UI designers want. Many (or all, maybe) of our UIs override these styles when they use this BaseAddress component... I think it's better for the component to use the right styles in the first place :)

seeker25 (approved):

severinbeauvais:
To prevent risk to the UIs, I recommend that the UIs do not import this latest version of sbc-common-components this week. (The only exception is Edit UI, which is the UI that this ticket was created for, and which I will review carefully.)

seeker25:
Sounds good, do we want to leave it unmerged until then?

severinbeauvais:
> Sounds good, do we want to leave it unmerged until then?

I was hoping to use these changes in Edit UI so I don't have to add CSS overrides there.

PR #330, opened 2023-07-27
CI/CD fixes - use the eslint from business-edit-ui + Migrate to Vite / Vitest

seeker25:


seeker25:
I was stuck scratching my head wondering why JEST wouldn't work, so I decided to just nuke it all and move to Vite / Vitest. This should give me some good practice to upgrade auth-web later.

seeker25:
@TMunro7 please fix BrowserVersionAlertView.vue to use the component - look at `MobileDeviceAlertView.vue`

PR #333, opened 2023-08-02
17195 - Pinia upgrade

seeker25:
Togo: account.ts - userStore issue - also minor warnings in unit tests

*Issue #:*
https://github.com/bcgov/entity/issues/17195


vue/sbc-common-components/src/store/account.ts
```diff
@@ -0,0 +1,122 @@
...
+        return this.currentUser
+      }
+    },
+    async getCurrentUserProfile (isAuth = false): Promise<UserProfile> {
+      try {
+        const response = await UserService.getUserProfile('@me')
+        const userProfile = response?.data || {} as UserProfile
+        if (isAuth) {
+          // const userStore = useStore('user') TODO: Figure this out!!!
+          // this.context.commit('user/setUserProfile', userProfile, { root: true })
```
seeker25, line 92: Last bit I have to figure out

vue/sbc-common-components/src/store/account.ts
```diff
@@ -0,0 +1,122 @@
...
+      if (response && response.data) {
+        const userProfile = response.data
+        // update the first name and last name for the users
+        const updateProfile:KCUserProfile = {
+          ...this.currentUser,
+          lastName: response.data.lastname,
+          firstName: userProfile.firstname
+        }
+        this.currentUser = updateProfile
+        return this.currentUser
```
seeker25, line 83: Seems weird to return something I set, but I wanted to be consistent with what was there before. I don't think the return for these values are used, but I'll have to check all of the projects.

vue/sbc-common-components/src/views/Home.vue
```diff
@@ -161,7 +140,6 @@
 <script lang="ts">
```
seeker25, line 140: This code is only used for testing, I'll rename it

vue/sbc-common-components/tests/unit/sbc.signout.spec.ts
```diff
@@ -1,10 +1,10 @@
 import { mount } from '@vue/test-utils'
 import SbcSignout from '@/components/SbcSignout.vue'
 import vuetify, { createVueRouter } from './setup'
-import { createStore } from 'vuex'
-import AccountModule from '@/store/modules/account'
-import AuthModule from '@/store/modules/auth'
 import { it, describe, expect, vi, beforeEach } from 'vitest'
+import { setActivePinia, createPinia } from 'pinia'
```
seeker25, line 5: These could probably get moved to a setup.ts

vue/sbc-common-components/src/components/SbcSystemBanner.vue
```diff
@@ -1,43 +1,35 @@
 <template>
   <v-alert
-    v-model="show"
+    v-model="state.show"
```
seeker25, line 3: Not sure if there is a better way of doing this with script setup

kialj876:
Hey Travis this looks great! Any reason you went with the options version for the stores? I'm just thinking if someone wanted to add in component hooks, composables, watchers, router stuff or injects from the app for anything they'd need to refactor it to setup I think

vue/sbc-common-components/src/components/SbcSystemBanner.vue
```diff
@@ -1,43 +1,35 @@
...
-  }
+const props = defineProps({
+  setDismissible: { default: false, type: Boolean },
+  setIcon: { default: 'mdi-information', type: String }, // See https://material.io/resources/icons/?style=baseline
+  setMessage: { default: '', type: String },
+  setShow: { default: false, type: Boolean },
+  setType: { default: 'warning', type: String }
+})
+
+const state = reactive({
```
kialj876, line 27: In general I'd just define these out of the state object for script setup components

vue/sbc-common-components/src/components/SbcSystemBanner.vue
```diff
@@ -1,43 +1,35 @@
...
-      ...state
-    }
-  }
+const props = defineProps({
+  setDismissible: { default: false, type: Boolean },
+  setIcon: { default: 'mdi-information', type: String }, // See https://material.io/resources/icons/?style=baseline
+  setMessage: { default: '', type: String },
+  setShow: { default: false, type: Boolean },
+  setType: { default: 'warning', type: String }
+})
```
kialj876, line 25: I would have this as 
```
const props = defineProps<{
dismissible: boolean,
icon: string,
message: string,
show: boolean,
type: string
}>()
```
It forces the parent to set them always (no defaults), but it is much nicer to read I think

vue/sbc-common-components/src/components/SbcSystemBanner.vue
kialj876: see my comment below -- should make it a lot nicer

vue/sbc-common-components/src/store/notification.ts
```diff
@@ -0,0 +1,65 @@
...
+  function markAsRead (): number {
+    const nl = JSON.parse(ConfigHelper.getFromSession(SessionStorageKeys.WhatsNew) || '{}')
+    nl.map(notification => { notification.read = true; return notification })
+    state.notifications = nl
+    state.notificationUnreadCount = nl.filter(notification => !notification.read).length
+    return state.notificationUnreadCount
+  }
+
+  return {
+    state,
```
kialj876, line 58: You might be able to dereference this like `...toRefs(state)` and that way people can access individual variables easier
i.e. 
```
import {useNotificationStore} from '@/store'
const { notifications } = useNotificationStore()
```
I'm not 100% sure this works but if it does I think that would be nicer for us

seeker25: I'll give it a shot

kialj876: Oh looks like you would need to do it like `const { notifications } = storeToRefs(useNotificationStore())` otherwise reactivity breaks

seeker25: Bugger lol back to what it was

PR #358, opened 2024-05-27
Bumped up version numbers in package.json and package-lock.json for the updated Header component. 

EPortman:
*Issue #:*
https://github.com/bcgov/entity/issues/<Put the github issue number here>

*Description of changes:*
 - Bumped up version numbers
 - Made it so case does not matter when passing in environment variables. Also made it so both 'dev' and 'development' can be passed in as an environment variable. 

### DIFF
--- a/vue/sbc-common-components/src/components/SbcHeader.vue
+++ b/vue/sbc-common-components/src/components/SbcHeader.vue
@@ -464,8 +464,9 @@ export default class SbcHeader extends Mixins(NavigationMixin) {
 
   @Watch
   get alertColor(): string {
-    switch (this.environment) {
+    switch (this.environment.toUpperCase()) {
       case 'DEV':
+      case 'DEVELOPMENT':
         return 'success';
       case 'TEST':
         return 'error';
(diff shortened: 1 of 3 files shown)


vue/sbc-common-components/src/components/SbcHeader.vue
```diff
@@ -464,8 +464,9 @@ export default class SbcHeader extends Mixins(NavigationMixin) {
 
   @Watch
   get alertColor(): string {
-    switch (this.environment) {
```
seeker25, line 467: this.environment?.toUpperCase()?

otherwise if this.environment is null, it will throw an exception

seeker25: nvm, we're good:

```
@Prop({ default: '' }) environment!: string;
```

seeker25 (approved):

seeker25:
Publishing on merge:

image

seeker25:
https://www.npmjs.com/package/sbc-common-components

EPortman:
Thanks Travis!

PR #368, opened 2024-06-13
DO NOT MERGE - this is a diff of what's changed since Release 3.0.8

severinbeauvais:
DO NOT MERGE


severinbeauvais:
Is this the same as the sum of individual releases from 3.0.8 until now?

https://github.com/bcgov/sbc-common-components/releases

vue/sbc-common-components/src/components/SbcHeader.vue
```diff
@@ -489,6 +489,7 @@ export default class SbcHeader extends Mixins(NavigationMixin) {
       // checking for account status
       await this.checkAccountStatus()
     }
+    this.$emit('account-data-loaded')
```
severinbeauvais, line 492: How do you see this being used? Would it update a store property that a UI could do something with (or await)?

PR #375, opened 2024-10-03
handle empty dict + empty list

bolyachevets:
*Description of changes:* 
This looks like the reason we've been getting 'null' values instead of '{}'


pwei1018 (approved):

seeker25:
@ochiu we're looking at incorporating this, should fix our `null` return problem hopefully

PR #390, opened 2025-05-08
update the footer and package version

phungchithanh:
update the footer and package version

*Issue #:*
https://github.com/bcgov/entity/issues/27546

*Description of changes:*
update the footer and package version


vue/sbc-common-components/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "sbc-common-components",
-  "version": "3.1.3",
+  "version": "3.1.4",
```
seeker25, line 3: 2nd time telling you, run `npm i` so package lock gets updated

phungchithanh: @seeker25 : I updated the package-lock.json file. thanks

PR #397, opened 2026-03-05
Fix SbcHeader Pinia store reactivity

seeker25:
Replace `beforeCreate` hook using `mapState`/`mapGetters`/`mapActions` with direct store access via computed getters and method wrappers
The `beforeCreate` pattern lost reactivity in Vue 2.7 + Pinia + Vite environments


ochiu (approved):

severinbeauvais (approved):
Let's try this.

